Issue #183, opened 2018-04-13
Surrogate Samples Aren't Auto-Generated


CSRussell2319:
Following the "Tutorial - Surrogates with UQ Tools" section in the FOQUS manual part 3 sub-parts b and c imply that by clicking done the "Flowsheet Results" area will fill with the samples. Instead the user needs to switch to the Uncertainty tab and launch the runs for them to appear. I also have an iReveal issue that may or may not be related. I've provided screenshots below. The first image is after attempting to add samples.

![surrogate samples issue 1](https://user-images.githubusercontent.com/37089145/38756305-4e39b8b6-3f26-11e8-9239-7df087426646.PNG)
![surrogate samples issue 2](https://user-images.githubusercontent.com/37089145/38756307-4e688fa6-3f26-11e8-8d9c-5742fa0b2a4f.PNG)
![surrogate samples issue 3](https://user-images.githubusercontent.com/37089145/38756309-4e7e5692-3f26-11e8-8f80-5525b4d5fecb.PNG)

fsoepyan:
@eslickj will look into this

ksbeattie:
Waiting for review/merge of #194 ...

fsoepyan:
Taken care of in PR #194.

Issue #186, opened 2018-04-16
Flowsheet: Results Table Mouseover


fsoepyan:
We would like the mouseover description for the "Results" table to say "Results and Filtering" instead of only "Results".

fsoepyan:
Taken care of in PR #194.

PR #194, opened 2018-04-19
Updated code for FOQUS

fsoepyan:
- I removed the old "surrogateFrame" and "surrogateFrame_UI" files
- I added new "surrogateFrame" and "surrogateFrame_UI" files such they no longer refer to iReveal (which is now a stand-alone tool)
- This should take care of one of the problems in Issue #183


fsoepyan:
- I deleted the old "uq_model" TEX file
- I added a new "uq_model" TEX file with additional steps
- This should clarify what the user needs to do, and take care of the other problem in Issue #183

fsoepyan:
- I deleted the "mainWindow" py file
- I added a new "mainWindow" py file
- This should take care of Issue #186

Issue #222, opened 2018-06-12
Problem saving/opening files with outdated Pandas


fsoepyan:
Miguel (@MAZamarripa) reported this following issue:
- The file did not save, and would not open.
- After trying to save the file, this error message appeared:

![save_06_12_2018](https://user-images.githubusercontent.com/30811140/41310365-4d087826-6e4f-11e8-802a-455425b513db.jpg)

fsoepyan:
I think once PR #229 is merged, we can close this issue.

fsoepyan:
PR #229 has been merged.

PR #229, opened 2018-06-19
Added pandas version requirement for setup.py

fsoepyan:
This change is also for issue #222

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ install_requires=[
     'scipy',
     'numpy',
     'cma',
-    'pandas==.23'],
+    'pandas>0.20'],
 
 #dependency_links=[]
 dependency_links=['git+https://github.com/CCSI-Toolset/turb_client@2.0.0-alpha#egg=TurbineClient']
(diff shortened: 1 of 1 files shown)


fsoepyan:
John (@eslickj), I'm not sure if this is correct, but can you please check? Thank you.

eslickj:
The ==0.23 should go inside the quotes at least (I think 'pandas'==.23 probably just evaluates to false).  I'm not really sure about the format, but I think that's right if you fix the quotes.  You'll have to run your version of setup.py to make sure it works as expected.

fsoepyan:
- I moved the ==.23 inside the quotes
- The previous code results in error, but the new code is fine, I think

pandas_06_19_2018

fsoepyan:
- I uninstalled Pandas using Anaconda by typing: pip uninstall pandas
- I had pandas-0.22.0
- Then, I reinstalled Pandas by typing: python setup.py develop
- The setup was successful
- I checked my version of Pandas, and I now have Pandas version 0.23.1
- This should take care of Issue #222
- I will test FOQUS and let you know what happens

fsoepyan:
- I managed to save and open a FOQUS file
- I think once this PR is merged, we can close issue #222

eslickj:
Looks good.

Issue #230, opened 2018-06-19
Add R installation details to INSTALL.md file


ksbeattie:
There are several packaged that are required to be installed in R for FOQUS.  Add details on how to install R and what to install inside of R.

fsoepyan:
I believe that once PR #233 is merged, we can close this issue.

ou3llnl:
#233 has been merged.

Issue #207, opened 2018-05-18
SolventFit: Error when Running SolventFit in "Basic Data" Tab


fsoepyan:
- I followed the instructions in Section 5 (SolventFit) of the user's manual.
- I got the following error:

<img width="960" alt="solventfit_04_19_2018" src="https://user-images.githubusercontent.com/30811140/40250451-5329789a-5aa4-11e8-8ce9-323bb40ea68e.png">

fsoepyan:
- John (@eslickj) and I modified line 836 in the file "foqus_lib\gui\solventfit\solventFitSetupFrame", where we added "()":

show=showList, rpath = self.parent().parent.dat.foqusSettings.rScriptPath,

- We did not get the previous error message

- Instead, we get this error message:

<img width="265" alt="solventfit_05_18_2018" src="https://user-images.githubusercontent.com/30811140/40252102-44bc7ee2-5aa9-11e8-85a2-c4925bdf49bd.png">

fsoepyan:
- Immediately after closing the 1st error message, we get this error message:

<img width="265" alt="solventfit2_05_18_2018" src="https://user-images.githubusercontent.com/30811140/40252266-da7aa6e8-5aa9-11e8-9eee-75fdd2d13c2d.png">

ou3llnl:
So this is a problem with additional R packages (abind and possibly others) not having been installed in your version of R. There are instructions in the manual. The solvent fit section will list the packages but will refer you to the acosso surrogate section for how to install those packages into your R.

ksbeattie:
@fsoepyan, says he will retest with the correct R packaged installed.

fsoepyan:
- I installed the following packages in R: 'quadprog', 'abind', and 'MCMCpack'
- At one point while running SolventFit, I got this error message:

<img width="960" alt="solventfit_06_20_2018" src="https://user-images.githubusercontent.com/30811140/41673299-8c1c2684-748a-11e8-9217-c4c01814cb28.png">

fsoepyan:
- I have just installed the 'MASS' package in R
- I will test SolventFit again

fsoepyan:
- I have these R packages installed: 'quadprog', 'abind', 'MCMCpack', and 'MASS'
- I reran SolventFit
- This time, I got this error message:

[...]

PR #233, opened 2018-06-20
INSTALL.md: Added instruction for installing R and packages

fsoepyan:
This change is to take care of some of the problems in Issue #207


ksbeattie:
I believe this also resolves #230

Issue #212, opened 2018-06-05
Bayesian Inference Exits Testing Scripts


CSRussell2319:
When attempting to perform automated testing for UQ Bayesian inference something triggers the script to finish prematurely. After clicking the "Infer" button a dialog box opens with code running then that box closes around the time of the `psuade>` mark in the picture below. That's when the rest of the testing script runs before continuing the inference script. Ideally this could all be run from the same window and the testing script could just wait for it to run. After PR #211 is merged this can be tested by running the test file from the console with this command:

`python foqus.py -s "examples\Smoke Tests\uq_smoke_test_inference.py"`

![b_inference](https://user-images.githubusercontent.com/37089145/40997947-84044c1a-68c3-11e8-903f-a8309bf26507.PNG)

CSRussell2319:
This also occurs in OUU analysis. I think it's directly tied to calling psuade since that seems to be the common thread here. With OUU I can actually close foqus with my script before the psuade script finishes running.

CSRussell2319:
Fixed the issue for Inference with this:
`MainWin.app.activeWindow().inference.inferencer.thread.terminate()`
For the InferenceDialog window.

I'll look into the OUU case to see if there's a similar solution available.

CSRussell2319:
Will close once PR #234 is merged

Issue #213, opened 2018-06-05
Add Testing Suite for UQ Functions


CSRussell2319:
Create a testing suite that runs through the different parts of the UQ tab to test different combinations of selection boxes.

PR #234, opened 2018-06-21
Inference and OUU Can Run and Close Automatically

CSRussell2319:
Fixed the issue where inference and ouu would finish the script before psuade finishes running. Also updated the inference smoke test to include error logging and multiple tests. Fixes issue #212 and finishes current development for #213.


CSRussell2319:
@ksbeattie I tried removing some of these functions to a separate file and it introduced a lot of issues. I might return to this later, but currently the default "MainWin" isn't passed to the imported script giving errors during run-time. At the moment having the extra lines of repeated code seems to be the easiest option.

PR #240, opened 2018-06-29
Added documentation header for each node.js module

boverhof:


ksbeattie:
@boverhof, was there an issue for this PR?

Issue #251, opened 2018-08-01
UQ: Samples in UQ do not show up in Flowsheet and Surrogates tables


fsoepyan:
- In the UQ tab, I would "Launch" the samples to obtain the values of the output variables for the sample points
- After FOQUS finishes this task, the values of the sample points do not appear in the Flowsheet table and the Surrogates table

eslickj:
I did a quick test of this and it works for me.  I fixed the pandas issue, but unless the sample didn't appear after saving and reloading, I don't think this is related to issue #250.

@fsoepyan can you provide steps to reproduce this?

fsoepyan:
- This issue is taken care of with pull request #256.
- I will close this issue.

Issue #250, opened 2018-08-01
Pandas Versions


fsoepyan:
There is a range of Pandas version, where if the version is too old or too new, FOQUS will not be able to work properly.

We should fix it for the newest version of Pandas.

ksbeattie:
@fsoepyan says that pandas 2.3 is too new.

ksbeattie:
@CSRussell2319 will look into which specific versions of pandas and matplotlib (see #227) we should use and make a pitch to the team for using those.

CSRussell2319:
@fsoepyan please try pandas 0.21 and report back.

fsoepyan:
- Pull request #256 has been accepted.
- This takes care of the issue, so I will close this issue.

Issue #237, opened 2018-06-28
Conda env standardization


ngbrenda:
To standardize the libraries that we are all using, can we create a requirement file and distribute it as part of the install:

`conda list --export > requirements.txt`

That way, all of us can easily duplicate the exact environment (let's call it "foqus_env"):

`conda create --name foqus_env --file requirements.txt`

CSRussell2319:
I've been making some progress on this in my fork. I currently have a working yml file, but I want to make sure all pieces of Foqus work in it before submitting a PR.

ksbeattie:
I've merged PR #247 but I'm not closing this issue just yet as I ran into the following problem when working under Linux:

When running `conda env create -f foqus_env.yml`
Had to comment out 'pywin32' from foqus_env.yml, as it appears that is not available on Linux?  Anyone know if there's a way to make this os-specific in the .yml file?

When running: `python foqus.py` got:
```
Traceback (most recent call last):
  File "foqus.py", line 48, in guiImport
    from PyQt5.QtWidgets import QApplication, QSplashScreen, QMessageBox, \
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```
But googling found that installing:
  `apt-get install libgl1-mesa-glx`
fixed that.  I got this idea from: https://github.com/conda-forge/pygridgen-feedstock/issues/10

Perhaps this is all known as an issue with running pyqt under linux?

CSRussell2319:
I ran my full suite of automated tests with the conda environment and had UQ fail for response surface visualization and uq analysis because matplotlib 1.5.3 doesn't recognize the facecolor property. I've attached the log files generated by my tests. We need to pick a version of matplotlib that fixes #227 and the issues I'm mentioning here.

[AutoErrLog_uq_response_surface_visualization.txt](https://github.com/CCSI-Toolset/FOQUS/files/2237345/AutoErrLog_uq_response_surface_visualization.txt)
[AutoErrLog_uq_response_uq_analysis.txt](https://github.com/CCSI-Toolset/FOQUS/files/2237349/AutoErrLog_uq_response_uq_analysis.txt)

Issue #227, opened 2018-06-13
Trouble Starting Optimization Using Automation


CSRussell2319:
I encounter an error when trying to perform the following:
`MainWin.optSetupFrame.optMonitorFrame.startButton.click()`

This is the error message I receive:
![start_opt_automation_issue](https://user-images.githubusercontent.com/37089145/41377709-29c60ace-6f1a-11e8-9c6c-9467bbdee757.PNG)

Is there an additional load/apply function that needs to be run before clicking this button? After the error I can click the button manually and it will run normally making me think I'm missing some in-between function.

eslickj:
@CSRussell2319, can you attach the script where this problem shows up?

CSRussell2319:
@eslickj the file is actually already in the master branch [here](https://github.com/CCSI-Toolset/FOQUS/blob/master/examples/Smoke%20Tests/optimization_smoke_test.py)

eslickj:
I was able to run the script without any problem.  I'm not sure at all but I'll guess at a solution. Try putting a call to go between the switching tab and the click on start.  It could be an issue with threads where there is not quite enough time to finish something before starting the optimization.  I really don't know if anything is multi-threaded here.  It's just a guess.  The optimization itself does run in a separate tread at least.  If adding the go doesn't work try using the sleep argument to have it sleep longer.  If that doesn't work let me know and I'll look at it harder.

CSRussell2319:
I've tried both of those solutions and I still get the same issue. What version of matplotlib do you have installed?

eslickj:
That's too bad.  This could be pretty hard to debug.  I have 1.5.3.  I can switch to your version and give it a shot too.

eslickj:
Oh and PyQt is at version 5.6.0 for me.

CSRussell2319:
@eslickj I downgraded matplotlib (and it's dependent packages) down to 1.5.3 via conda and now optimization works via my script. It's probably a problem with matplotlib's new libraries. I'm running the latest version of all anaconda packages. Below are the packages that conda downgraded when I installed the older version of matplotlib:

krb5:         1.13.2-0              --> 1.16.1-hb4d044e_6
matplotlib:   1.5.3-np111py27_1     --> 2.2.2-py27h8803d4e_1
numpy:        1.11.3-py27hfef472a_4 --> 1.14.5-py27h911edcf_0
pillow:       4.2.1-py27hbae4915_0  --> 5.1.0-py27h901f87c_0
[...]

PR #255, opened 2018-08-08
Updated Conda Environment and Added Testing Scripts

CSRussell2319:
Updated yml file that has versions of matplotlib and pandas that should satisfy #250, #227, and #251. This is progress towards #237. The other scripts are useful for creating, activating, using, deactivating, and destroying a conda environment on a unix environment (windows users can use the bash shell in powershell and it works fine).


CSRussell2319:
I've moved the testing scripts to the test/auto_test directory with matching pointers in the script. I've left the yml file in the main directory since I think the intent is to use it generally in addition to automated testing. Sorry for the long list of commits, I didn't realize all of the commits in my fork would be carried over.

PR #258, opened 2018-08-10
Updated Turbine configuration: FOQUS considers Aspen simulation successful even if there is error in Aspen

fsoepyan:
I updated turbineConfiguration.py, such that FOQUS will consider an Aspen Plus simulation to be successful, even though the Aspen Plus simulation reports to the user that the simulation completed with warnings and/or errors.

### DIFF
--- a/foqus_lib/framework/sim/turbineConfiguration.py
+++ b/foqus_lib/framework/sim/turbineConfiguration.py
@@ -1129,8 +1129,8 @@ class turbineConfiguration():
         comProb = False
         res = None
         state = 'submit' #initial state of the job
-        failedStates = ['error', 'expired', 'cancel', 'terminate']
-        succesStates = ['success', 'warning']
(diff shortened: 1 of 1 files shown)


eslickj:
I don't think we want to do this.  If the simulation ends in an error you want to know, and don't want FOQUS to say it was successful.  At least this should be optional.  This also doesn't fix the issue of results not being returned for outputs when there is an error, and warnings being mistaken for errors.

Let's hold off on this until we track down the main source of the problem.

ksbeattie:
@eslickj says this should be addressed via #268, not this way.

fsoepyan:
- @boverhof has fixed Issue #268 by updating TurbineLite
- Should we reject this PR and close it?

ksbeattie:
Closing this as it was fixed by updating TurbineLite (https://github.com/CCSI-Toolset/turb_sci_gate/releases/tag/2.0.1).  See issue #268

PR #260, opened 2018-08-16
Added reference to README

fsoepyan:
- I added a reference in the README file
- The reference is the Miller et al. (2017) paper, which is a book chapter


fsoepyan:
- I will merge this PR since Keith (@ksbeattie) has approved it
- I think the reference is fine, as it describes the various features of FOQUS (e.g., Optimization, UQ, OUU, etc.) and describes how FOQUS works

PR #273, opened 2018-08-31
Changed the link for the TurbineLite installer

fsoepyan:
The link for the TurbineLite installer now goes to the August release page


ksbeattie:
I'm closing this without merging it, not because it is incorrect but because I've started another PR (#274) with this (and other things) already in it.  Sorry, I should have look for this here first!

PR #281, opened 2018-09-12
AWS Gateway Integration

boverhof:
Mostly changes to various GET/PUT/POST operations and the foqus_service. 

Changes to foqus and foqus_lib mostly changing resource names to "LOWERCASE" because AWS is discerning..


fsoepyan:
- I went to the "Settings" tab, and under "FOQUS", I changed the "FOQUS Flowsheet Run Method" to "Remote"
- I went back to the Flowsheet, and whenever I select any of the items under "Turbine" in the "Node Editor", I get an error message that looks like this:

turbine_09_18_2018

- I also tried to upload a simulation into Turbine, with the following options:

turbine2_09_18_2018

- I got the following error message:

turbine3_09_18_2018

- @boverhof, is there anything I need to do in the Anaconda prompt to resolve these issues?

Issue #263, opened 2018-08-20
UQ: Problem with loading samples in UQ model from CSV file


puttak:
Hi,
I am getting error when i am trying load a csv file format data samples in UQ mode.

fsoepyan:
Putta (@puttak), can you please provide a screenshot of your error, and a few short bullet points on how to reproduce the error?

puttak:
![csv issue](https://user-images.githubusercontent.com/30050105/44367151-f6f67300-a49c-11e8-871b-02ac579500a5.jpg)

sotorrio1:
@puttak Please test now and let me know if it is fixed so we can close the issue. Thanks!

ksbeattie:
@fsoepyan will contact @puttak to see if he's still working on the project a/o can help confirm this.

ksbeattie:
@puttak will confirm this

ksbeattie:
@fsoepyan has reached out to him, @puttak is in Norway and said he could look at it.  @fsoepyan will try testing it himself to save time.

fsoepyan:
- I opened the "Rosenbrock_no_vectors" example from under the "examples/UQ" folder
- I went to the UQ tab
- I created a CSV file, saved it, and selected "CSV (MS-DOS)" for the file type (following the suggestion of @CSRussell2319)
- I managed to upload the CSV file successfully (Thank you!)
- I will close this issue

PR #282, opened 2018-09-17
Fixes issue #263. Error uploading csv files in UQ

sotorrio1:
@puttak could you test if this is no longer an issue? Thanks!


fsoepyan:
- I have a CSV file in Excel:

csv1_11_07_2018

- I opened the CSV file in Notepad++:

csv2_11_07_2018

- When I clicked "Load from File", and selected the file, I got an error message:

csv3_11_07_2018

sotorrio1:
@fsoepyan , that's different from the error that @puttak got. It looks like there is something wrong with the format of your csv file. I'll send you a different csv file in an email and you can try with that one.

Issue #297, opened 2018-10-10
Optimization Run ends in Assertion Error "nodeError"


boverhof:
## Description
Assertion error on "nodeError".  NOTE this could be a duplicate bug report..

## Traceback
```
['Traceback (most recent call last):\n',
'  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\foqus-2017.12.0.0-py2.7.egg\\foqus_lib\\framework\\optimizer\\optimization.py", line 76, in run
    self.optimize()',
'  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\foqus-2017.12.0.0-py2.7.egg\\foqus_lib\\framework\\optimizer\\BFGS.py", line 188, in optimize options={\'ftol\':ftol, \'eps\':eps, \'maxfun\':maxeval})',
'  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\scipy\\optimize\\_minimize.py", line 450, in minimize
    \n    callback=callback, **options)\n',
    '  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\scipy\\optimize\\lbfgsb.py", line 328, in _minimize_lbfgsb\n    f, g = func_and_grad(x)\n',
    '  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\scipy\\optimize\\lbfgsb.py", line 274, in func_and_grad\n    g = _approx_fprime_helper(x, fun, epsilon, args=args, f0=f)\n',
    '  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\scipy\\optimize\\optimize.py", line 628, in _approx_fprime_helper\n    grad[k] = (f(*((xk + d,) + args)) - f0) / d[k]\n',
    '  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\scipy\\optimize\\optimize.py", line 292, in function_wrapper\n    return function(*(wrapper_args + args))\n',
    '  File "C:\\ProgramData\\Anaconda2\\lib\\site-packages\\foqus-2017.12.0.0-py2.7.egg\\foqus_lib\\framework\\optimizer\\BFGS.py", line 113, in f\n    objValues, cv, pv = self.prob.runSamples([x], self)\n',
[...]

PR #298, opened 2018-10-10
Fixing add result so fis something is missing FOQUS won't throw exceptrion  

eslickj:
Fixing problem where FOQUS would kill and optimization or UQ run if a job came back with incomplete results.


eslickj:
This partly addresses issue #297

boverhof:
I'm going to accept this, but I think we should at least have "log" debug messages for situations that are unexpected or seemingly will cause problems.

eslickj:
I added a log message to graph.py where it loads the results dictionary to log an error if there is no inputs field.  The inputs field is one thing that should always be there.  If it's missing it is probably a sign that something crashed or timed-out or something. The if the FOQUS consumer returns something it should always have input.

PR #306, opened 2018-10-17
Optimization Smoke Tests

CSRussell2319:
Added many smoke tests to test each solver and each NLopt solver. Added all of these tests to the auto test script and added a section in the auto test script for long tests that take a very long time to run.


boverhof:
@CSRussell2319 I added you to a pull request for "optimizer_bfgs_s3_test"

https://github.com/CCSI-Toolset/FOQUS/pull/305

I'm wondering if the test here
```
examples/Smoke Tests/optimization_smoke_test_BFGS.py
```
Can be integrated with the one I've developed..

CSRussell2319:
@boverhof I don't think the tests can be integrated. My test goes through the GUI and triggers a problem using the BFGS solver while yours appears to test only the back end of the process. I can share the problem I'm using but I think these two tests serve different purposes.

Issue #347, opened 2019-01-23
Adding Dependencies in setup.py


eslickj:
Previously, on windows installing PyQt with pip on Windows would cause result in getting the source packages.  There would be an attempt to compile, which would always fail.  It seems like binaries an now available, and with our switch to Python 3, we should finally be able to add pyqt5 and have all the dependencies handled there.

eslickj:
See PR #348

Issue #342, opened 2019-01-22
Update install.md doc with python 3 details


ksbeattie:
Now that py3 is in master, this needs updating.

eslickj:
See PR #348

PR #348, opened 2019-01-23
Updated Install/Instructions

eslickj:
I updated the install instructions and the setup.py file.  In setup.py I tried to get it to pick up all the data files so there is no need to do the "develop" option.  Setup.py also now includes the PyQt stuff.  In Windows, when you tried to pip install pyqt it would fail trying to compile Qt, but I think they have precompiled versions now days.

Assuming you have Python 3, you should just need to do ``python setup.py install`` as long as you have Python >= 3.6 or so.  I wrote the instructions still with an optional conda environment set up.

If you have any trouble with the instructions or the installer, we can solve them before merging this or merge then fix it up.  Whatever is easiest for everyone is fine with me.


eslickj:
Issues #342 and #347

Issue #312, opened 2018-10-18
Sorting Fails with an Operation


CSRussell2319:
If I attempt to sort flowsheet results with two conditions it will fail with an error for every operation except "NOT". I have tested the rules individually and they run without error. Below is the picture of the error. I have included screenshots of the error and a report of the output from my test.

![sort_error_1](https://user-images.githubusercontent.com/37089145/47174703-933acb80-d2ce-11e8-8b4c-a65b37b592b9.PNG)
![sort_error_2](https://user-images.githubusercontent.com/37089145/47174704-93d36200-d2ce-11e8-9ee8-d92dafe05fc4.PNG)
[AutoErrLog_fs.txt](https://github.com/CCSI-Toolset/FOQUS/files/2493089/AutoErrLog_fs.txt)

eslickj:
This one is user error.  It's partly my fault though for writing this thing using reverse Polish notation.  The and pops two things off the stack and put one back on.  In the example screen shot, and is in the middle.  There is only one thing to on the stack at that point. 

One thing we could do to get around this is finish removing this filter system and use calculated columns to do it instead.  You could pick a calculated column that evaluates to True or False.  True is in False is out.  The syntax for calculated columns is just regular Python so it would be even more flexible and easier to understand.

eslickj:
Change the way this works with PR #353.  That may get refined a bit, but I don't think we'll go back to this way of filtering.

PR #353, opened 2019-01-24
Change Filter Method to Use Calculated Columns

eslickj:
This is an update to make filtering easier to understand.  We will need to update the documentation at some point though.  This could break some of @CSRussell2319's tests if there are filter tests.

This is related to issue #312 

The box for a filter now takes a Python expression.  This expression should return a numpy array or list that has the same number of elements as rows in the data table where each entry is true or false.  True to include a row false to exclude it.  The syntax is the same as calculated columns.  There is a c("my_column") function that returns a numpy array of values in my_column. You can write your Python expression how you want, but I recommend numpy logic functions https://docs.scipy.org/doc/numpy-1.15.1/reference/routines.logic.html.  Numpy is imported as np.

Examples  using Rosenbrock_no_vectors.foqus.

No errors and x1 > 0:

```python
np.logical_and(c("input.Rosenbrock.x1") > 0, c("err") == 0)
```
x2 > 0 and x1 > 0:

```python
np.logical_and(c("input.Rosenbrock.x1") > 0, c("input.Rosenbrock.x2") > 0)
```
x2 > 0 or x1 > 0:

```python
np.logical_or(c("input.Rosenbrock.x1") > 0, c("input.Rosenbrock.x2") > 0)
```

If you are careful with parentheses and make sure the arrays you are doing logical operations on contain strictly boolean values you can still use Python's binary operators, but be careful it's easy to get an exception or a strange result.  If you get an exception you can just go back and edit the filter.  The "and" and "or" operators will cause Numpy to raise and exception about "the truth value of an array with more than one element," so if you see that you know what you did.  These examples also work but are more error prone and less recommended.

 No errors and x1 > 0:

```python
(c("input.Rosenbrock.x1") > 0) & (c("err") == 0)
```

x2 > 0 and x1 > 0:

```python
(c("input.Rosenbrock.x1") > 0) & c("input.Rosenbrock.x2") > 0)
```
x2 > 0 or x1 > 0:

```python
(c("input.Rosenbrock.x1") > 0) | (c("input.Rosenbrock.x2") > 0)
```

### DIFF
--- a/foqus_lib/framework/sampleResults/results.py
+++ b/foqus_lib/framework/sampleResults/results.py
@@ -17,67 +17,25 @@ import time
 from collections import OrderedDict
 
 class dataFilter(object):
-    DF_NOT = 0
-    DF_AND = 1
-    DF_OR = 2
-    DF_XOR = 3
-    DF_END_OP = 9
-    DF_RULE = 10
 
-    def __init__(self):
-        self.fstack = []
-        self.sortTerm = None
+    def __init__(self, no_results=False, lock=False):
+        self.filterTerm = None       # list of columns to filter by
+        self.sortTerm = None         # list of columns to sort by
+        self.no_results = no_results # if true return no matches
+        self.lock = lock
 
     def saveDict(self):
-        sd = {
-            'fstack':[],
-            'sortTerm':self.sortTerm}
-        for i in self.fstack:
-            if i[0] < self.DF_END_OP:
-                sd['fstack'].append(i)
-            elif i[0] == self.DF_RULE:
-                sd['fstack'].append([i[0], i[1].saveDict()])
+        sd = {'filterTerm':self.filterTerm,
+              'sortTerm':self.sortTerm,
+              'self.no_results':self.no_results,
+              'self.lock':self.lock}
         return sd
 
     def loadDict(self, sd):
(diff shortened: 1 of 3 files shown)


fsoepyan:
- I used the "examples/UQ/Rosenbrock_no_vectors" FOQUS file.
- I went to UQ, and generated and launched 50 samples successfully.
- I went back to Flowsheet, and went to the Data Table.
- When I clicked "Add/Edit Filters", a box appears:

calccol_01_24_2019

- @eslickj, can you please let me know what I need to do if I want to filter my results for outputs >= 10 and err = 0? What I typed in did not work.

eslickj:
You don't put the calculation there.  You add a calculated column and put the name of the column there.  You can add a list of columns and they get anded together, but I don't know how useful that is.  I know it's not clear since there wasn't actually any documentation.  Can you look at the tooltips and let me know if there is a way I can improve them?

eslickj:
@fsoepyan, I guess the way I did this was easy for me, but still probably confusing.  Let me change this so instead of using calculated columns, you can just put a Python expression right in the filer box, and I'll change the name of that box to "Filter Expression."

eslickj:
Okay, I changed it so you can directly put in filter expressions.  I'm going to revise the pull request description, so see there for more information.

Issue #310, opened 2018-10-18
Generate Samples Produces Error


CSRussell2319:
When in the optimization tab clicking the "Generate Samples" button produces an error. If this is an intentional error we should fix it to say what the user is doing incorrectly.

![image](https://user-images.githubusercontent.com/37089145/47167283-54e7e100-d2bb-11e8-871e-16cb5e7ecc0a.png)

PR #354, opened 2019-01-24
Fixed optimization generate samples dialog.

eslickj:
Parent argument was missing from dialog box init function.  Issue #310

### DIFF
--- a/foqus_lib/framework/optimizer/problem.py
+++ b/foqus_lib/framework/optimizer/problem.py
@@ -142,7 +142,7 @@ class problem():
                 self.solverOptions[dkey][okey] = ov
 
     def loadSamples(self, fname):
-        with open(fname, 'rb') as f:
+        with open(fname, 'r') as f:
             cr = csv.reader(f)
             head = [h.strip() for h in next(cr)]
             for h in head:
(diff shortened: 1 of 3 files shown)


fsoepyan:
- I tested this PR using the "examples/UQ/Rosenbrock_no_vectors" FOQUS file
- In Optimization, I set x2 and x3 as samples
- When I tried to generate samples, I got this error message:

opt_01_24_2019

eslickj:
@fsoepyan, I think what happened here is that you  tried you do a full factorial sampling and entered single values.  The input for those is lists.  I guess we probably ought to have some error checking and some extra information on the page.  Will add.

It works like this: x1 = [2, 4], x2 = [1,3] samples = [[2,1], [2,3], [4,1], [4,3]]

eslickj:
I added a check to catch the error @fsoepyan encountered that tells you exactly what you did wrong, and how to fix it.

eslickj:
I'll fix this one here too.  It's a problem left over from the Python 3 conversion.

eslickj:
@fsoepyan, I fixed and tested the read from file problem.  I'll merge this.  If you have any other problems go ahead and submit new issues.

PR #366, opened 2019-01-30
Batch File and requirements.txt

eslickj:
I added requirements.txt which I hope will fix the weird import errors that we were getting with requirements installed from setup.py.  Installing those added a step to the install instructions.

I also took a shot at generating a batch file on windows at the end of setup.py.  It works good for me, but several people will probably need to try it to ensure it works for different setups.  I think as long as it looks reasonable it's probably fine to merge then test.

I also dropped the enumerated type value that was causing problems  with adodbapi.  I don't know why sometimes the enum is there sometimes and not others.  Maybe if has something to do with where you get your pacakages, but I replaced it with the number to avoid problems.

@boverhof, I added python-dateutil==2.5.0 to requirements because I have a problem where the latest version makes TurbineClient complain about the version being to old.  They may have removed something you are using.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,7 +12,8 @@ you can skip steps 1 to 4.
 3. Set up a foqus environment.
   > conda create -n foqus -python=3
 4. Activate the environment
-  > source activate foqus
+  - Linux: > source activate foqus
+  - Windows: > conda activate foqus
 
 If you create an environment in which to install FOQUS, you will need to ensure that environment is active before running FOQUS.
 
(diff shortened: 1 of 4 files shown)


eslickj:
I accidentally branched PR #367 off this so it got merged when that did

CSRussell2319:
The batch file seemed to be missing an ending quotation when I tried using it. Also, it might be better to use /K instead of /C since the user can't see any error messages from launching foqus if it closes automatically (I had an error about tqdm not being installed that I had to manually check since the cmd window closed before I could read the error). This could be done by creating an additional file labeled "Debug" that has the /K included. You removed the option to install turbine via ssh, which is needed for my corporate proxy to work. The batch file should be added to the git ignore list if it hasn't been already.

ksbeattie:
@CSRussell2319, seems like your suggestion might be worth following up on.  Is it captured in another issue or PR?

CSRussell2319:
No, I haven't opened any other issue to address this

ksbeattie:
I created issue #388 for us to follow up on, if desired.

Issue #378, opened 2019-02-08
SDoE Test Fails & Issue with Specification for Design Size 


jmorgan29:
When running an example, using the provided candidate files 'aggregate_candidates.csv' and 'aggregate_history.csv', and attempting to test the SDOE, the following error results:

2019-02-08 10:58:47,557 - CRITICAL - foqus.__main__ - unhandled exception
Traceback (most recent call last):
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\foqus-3.0.0-py3.6.egg\foqus_lib\gui\sdoe\sdoeAnalysisDialog.py", line 378, in testSdoe
    runtime = sdoe.run(self.writeConfigFile(test=True), self.maxDesignSize_spin.value(), test=True)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\foqus-3.0.0-py3.6.egg\foqus_lib\gui\sdoe\sdoeAnalysisDialog.py", line 315, in writeConfigFile
    os.makedirs(outdir, exist_ok=False)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\os.py", line 220, in makedirs
    mkdir(name, mode)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\jmorgan\\Documents\\FOQUS\\SDOE_files\\2019-02-08T10:58:47.557000'

Also, the Max Design Size cannot be set to a value higher than 3.

sotorrio1:
@jmorgan29 I just created PR #379 that should take care of the Windows specific issue when creating new solution directories. 

Regarding the issue about Max Design Size:
The candidate and history files you should use are the ones I added in PR #369 called respectively 'candidate.csv' and 'historical_data.csv'. SDOE creates 'aggregate_candidates.csv' and 'aggregate_history.csv' in the working directory under the SDOE_files directory. This capability exists in case the user wants to use more than one candidate file. If you haven't downloaded PR #369 just yet, you can find these same files under 'foqus_lib/framework/sdoe/'. 
I just tested the code in iOS and Windows and there is no limit in the Max Design Size. Please test again and let me know if you have any questions. Thanks!

jmorgan29:
[...]

PR #379, opened 2019-02-08
fixing issue with creating directories in Windows OS

sotorrio1:
Small change in the datetime format to create directories since it was causing some issues in Windows OS. Should take care of issue #378


sotorrio1:
@jmorgan29 if you feel comfortable reviewing and approving this merging feel free to do it. Thanks!

Issue #176, opened 2018-03-29
Basic Data Window is Too Small


CSRussell2319:
When using SolventFit the UI window doesn't show all input boxes by default. There's also no side bar to indicate that there is more to the window if expanded. User currently needs to resize the window to be able to see all of the options for this tool. The window in the manual's screenshot (pg 144) is perfectly sized.

Current Default:
![basic_data_window_too_small](https://user-images.githubusercontent.com/37089145/38112027-eeebb554-335d-11e8-813d-d2c3116bda95.PNG)
Expanded to See Everything:
![basic_data_window_see_all](https://user-images.githubusercontent.com/37089145/38112136-335387b2-335e-11e8-9fb9-3729763259b2.PNG)
Manual Screenshot:
![basic_data_window_manual](https://user-images.githubusercontent.com/37089145/38112382-0120b14c-335f-11e8-8a95-ecec9a287c9a.jpg)

ngbrenda:
I am having the same issue too (running on Mac). Is there a way to add scrollbar?

sotorrio1:
Added a scroll bar and testing it. Hopefully I can create a PR today to solve this issue.

sotorrio1:
@CSRussell2319 PR #380 should fix this issue in Windows and also in iOS. Please test and let me know :)
Thanks!

ksbeattie:
@sotorrio1 is waiting on #408 on if this will even need to be done.

sotorrio1:
PR #474

ksbeattie:
Closing this for now, possibly re-opening it if we want to keep the basic data tab.

PR #380, opened 2019-02-09
Basic data gui#176

sotorrio1:
Adding scroll area to Basic Data window so the user can see everything. It should take care of issue #176


sotorrio1:
@CSRussell2319 thanks for the feedback. I'll try to fix that as soon as possible :)

sotorrio1:
Closing this for now. I need to start over with some GUI changes. I will open a new PR when ready

PR #386, opened 2019-02-13
Adding a new sphinx make.bat file for windows.

eslickj:
This just adds a make.bat script, which seems to be the standard way to build sphinx docs on windows.  

Built docs are in: docs/build/html/index.html

@fsoepyan, once this is in, would you make sure you can build the docs and give them a read through and see if you can pick out what needs to be updated.  There are also some possible problems with references and enumerated lists from the conversion too we'll need to look out for.

### DIFF
--- a/docs/make.bat
+++ b/docs/make.bat
@@ -1,0 +1,35 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation
+
+if "%SPHINXBUILD%" == "" (
+	set SPHINXBUILD=sphinx-build
+)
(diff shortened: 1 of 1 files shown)


fsoepyan:
@eslickj: Thank you for the PR.

Do you have any recommendation for a software I can use to open RST type files?

Thank you!

ksbeattie:
@fsoepyan, you can edit .rst files in pretty much any editor.  It's just text, a lot like markdown.

A lot of IDEs these days probably have special modes for editing them and either displaying them or previewing them as they would appear in HTML.

Issue #396, opened 2019-02-19
FOQUS cannot locate files


jmorgan29:
I am including two issues here which seem to be related.

In the SDOE tab, I generate a space filling design, and then click 'Plot SDOE', and the following results:

2019-02-19 10:14:39,641 - CRITICAL - foqus.__main__ - unhandled exception
Traceback (most recent call last):
  File "c:\users\jmorgan\desktop\foqus-master - **2\foqus_lib\gui\sdoe\sdoePreview.py",** line 113, in plotSdoe
    sdoe.plot(fname, show=show)
  File "c:\users\jmorgan\desktop\foqus-master - 2\foqus_lib\framework\sdoe\sdoe.py", line 68, in plot
    df = load(fname)
  File "c:\users\jmorgan\desktop\foqus-master - 2\foqus_lib\framework\sdoe\df_utils.py", line 11, in load
    df = pd.read_csv(fname)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\parsers.py", line 678, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\parsers.py", line 440, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\parsers.py", line 787, in __init__
    self._make_engine(self.engine)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\parsers.py", line 1014, in _make_engine
    self._engine = CParserWrapper(self.f, **self.options)
  File "C:\Users\jmorgan\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\io\parsers.py", line 1708, in __init__
    self._reader = parsers.TextReader(src, **kwds)
  File "pandas\_libs\parsers.pyx", line 384, in pandas._libs.parsers.TextReader.__cinit__
  File "pandas\_libs\parsers.pyx", line 695, in pandas._libs.parsers.TextReader._setup_parser_source
FileNotFoundError: File b'C:Users\\jmorgan\\Documents\\FOQUS\\SDOE_files\\20190219T101433\\candidates_d4_n1000_w+G+lldg+L' does not exist****

[...]

PR #397, opened 2019-02-19
Sdoe locate files

sotorrio1:
Potentially fixes issue #396

### DIFF
--- a/foqus_lib/framework/sdoe/df_utils.py
+++ b/foqus_lib/framework/sdoe/df_utils.py
@@ -1,3 +1,4 @@
+import os
 import pandas as pd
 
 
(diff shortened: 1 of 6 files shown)


foqus_lib/framework/sdoe/df_utils.py
```diff
@@ -8,7 +9,7 @@ def write(fname, df):
 
 def load(fname):
     # load file as data frame
-    df = pd.read_csv(fname)
+    df = pd.read_csv(os.path.join('c:/', fname))
```
ksbeattie: This looks OS-specific.  What will happen here on a non-Windows platform?

sotorrio1: That was my first thought, but I tested it on Mac and still works...

jmorgan29: I verified that this fixed the SDOE related aspect of Issue #396

sotorrio1:
@jmorgan29 Let me try to find out the best way to fix this and I'll create a new PR. Glad to hear this little hack works...

PR #403, opened 2019-02-21
Updated LocalExecutionModule.py, line 314

fsoepyan:
@sotorrio1 spotted this error in line 314, and after the fix, I can successfully upload a csv file to the UQ module. Thank you!


fsoepyan:
This fix is for the py2 branch of FOQUS.

fsoepyan:
Thank you!

PR #419, opened 2019-03-05
Update INSTALL.md

eslickj:
I completely removed Miniconda to test the install again, and when I created a new environment I found that pip wasn't installed, so I added pip to the environment create line in the instructions.


sotorrio1:
When I create my conda environment, pip gets installed... but I guess it doesn't hurt to make sure that pip gets installed.

PR #429, opened 2019-03-07
Updated installation instruction to show how to open FOQUS

fsoepyan:


fsoepyan:
- Yesterday, when I tried to open FOQUS using "foqus.bat", it won't open, and I had to type "conda activate foqus" and "python foqus.py" to open FOQUS.
- For some reason, today, when I tried to open FOQUS using "foqus.bat", it opened successfully.

eslickj:
I'm going to close this PR.  If you have trouble with the bat file, open an issue.  Let's try to keep the instructions simple an consistent and fix problems rather than changing the instructions.  I may be worth adding this sort of information to a trouble shooting section though.  To only be looked at if things go wrong.

Issue #161, opened 2018-03-22
Option for SSH Download of Dependencies


CSRussell2319:
When running `python setup.py install` the installer looks for turb_client and attempts to install it. It uses an https link for downloads and installs. For my current network configuration that doesn't work and I need to use ssh instead. Is this a common enough situation that we should have a standard way to deal with it (maybe user input of how they would like to download dependencies)?

For my network I have to change the following:
setup.py line 53
Old:
`dependency_links=['git+https://github.com/CCSI-Toolset/turb_client@2.0.0-alpha#egg=TurbineClient']`
New:
`dependency_links=['git+ssh://git@github.com/CCSI-Toolset/turb_client@2.0.0-alpha#egg=TurbineClient']`

CSRussell2319:
#247 adds in the code to easily enable ssh download of dependencies, but we should document this. SSH is still not the default behavior and the only way to use ssh with this code is to manually comment the https line and uncomment the ssh line

ksbeattie:
I'd think that the most correct way of installing turb_client should be via a `pip install` and/or `conda install`.  That should fix this, right?  @boverhof, thoughts?

I'm going to put this on the Nov release board, so we can follow this idea....

CSRussell2319:
Yes, I believe that would fix this issue.

CSRussell2319:
Tried using pip to install the turbine client and received the same error message. 
Here's my code for the yaml file: `"--editable=git+https://github.com/CCSI-Toolset/turb_client@2.0.0-alpha#egg=TurbineClient"`
Looks like pip just does a git clone so it's no different than what the setup file does natively. If we could get the repo added to a pip or conda repo that might fix the issue.

ksbeattie:
Fixed with PR #283

Issue #435, opened 2019-03-08
Require git client?


eslickj:
The installation instructions say we need a git client.  That may be true for installing Tubine Client off github, but if we get the non-github pip install, we can probably drop the git client requirement from the install instructions.  I'm actually not really sure if you need the git client now.  If we can lose that requirement before the release the install will seem a bit simpler.

ksbeattie:
I agree we should drop that requirement.

This issue depends on: https://github.com/CCSI-Toolset/turb_client/issues/4

CSRussell2319:
I agree with the exception that we need to have developer installation/setup instructions somewhere. By removing things like this I, or others in my position, need to explain how to properly use this repository with Git. Other than our developer meetings there is little discussion of how we want newcomers to make contributions (branches, forks, etc) or how we expect changes to be tested using automated testing.

ksbeattie:
I'll remove this section from the install.rst doc.

PR #448, opened 2019-03-13
Use the pypi package for TurbineClient

ksbeattie:
Not a big change in code, but should simplify a few things.

Related to https://github.com/CCSI-Toolset/turb_client/issues/4 as well as https://github.com/CCSI-Toolset/FOQUS/issues/161, I believe.


ksbeattie:
Yes @eslickj, I believe that's already been done (at least commented out, for ref when the dev page is written) in #443, for #435

PR #449, opened 2019-03-13
Improve tear solve methods

eslickj:
Can wait until after release to merge.

This merges the Direct and Wegstein methods and increases logging output.  I also fixed a few minor issues, like direct not using the "fraction of range" convergence criteria.


eslickj:
@ksbeattie and @fsoepyan, I'm not sure if it would be better to wait until after the release to merge this, but since it fixes some bugs and seems to work for @fsoepyan and I, maybe if its not too late we should merge it.  I'm pretty comfortable with it, but I know it hasn't been extensively used yet. Either way is okay.

Issue #462, opened 2019-04-02
Plugin loading fails to encoding(?) issue on AppVeyor


perrenyang:
I've been trying to get FOQUS installed/built and running on AppVeyor for automated testing on a Windows platform. When the application attempts to launch, an error occurs during plugin loading:

```
2019-04-01 19:46:43,144 - DEBUG - foqus.foqus_lib.framework.session.session - Initializing session, Log file: C:\projects\tests\logs\foqus.log, Position: 0
2019-04-01 19:46:43,144 - INFO - foqus.foqus_lib.framework.plugins.pluginSearch - Loading Plugin: C:\projects\foqus-37w98\foqus_lib\framework\surrogate\ACOSSO.py
2019-04-01 19:46:43,144 - CRITICAL - foqus.__main__ - unhandled exception
Traceback (most recent call last):
  File "foqus.py", line 398, in <module>
    dat = session(useCurrentWorkingDir=True)
  File "C:\projects\foqus-37w98\foqus_lib\framework\session\session.py", line 239, in __init__
    self.loadPlugins()
  File "C:\projects\foqus-37w98\foqus_lib\framework\session\session.py", line 343, in loadPlugins
    charLimit = 150)
  File "C:\projects\foqus-37w98\foqus_lib\framework\plugins\pluginSearch.py", line 28, in __init__
    self.importPlugins()
  File "C:\projects\foqus-37w98\foqus_lib\framework\plugins\pluginSearch.py", line 42, in importPlugins
    l = self.idString in f.read(self.charLimit)
  File "C:\Miniconda3\envs\foqus\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 2835: character maps to <undefined>
Command exited with code 1
```
More complete context can be seen in the image below:
![image (1)](https://user-images.githubusercontent.com/44816826/55441605-38d1b680-5560-11e9-9119-776a81713427.png)

For comparison, when I run the same test on my own Windows machine, I see the GUI pop open and move through several different actions before exiting, leaving the following output:

![image](https://user-images.githubusercontent.com/44816826/55441668-720a2680-5560-11e9-9683-5939b4c5f255.png)

(probably you'll need to click on the images to load the full resolution)

eslickj:
[...]

PR #464, opened 2019-04-06
Fix plugin search

eslickj:
This fixes issue #462 

Added try/except and logging for finding future problems.   The problem was a 0x90 in ACOSSO.py.  I don't know how it got there, but I changed the plugin search code to read the files with UTF-8 encoding.  That fixed it.  That character must have gotten in recently somehow.

### DIFF
--- a/foqus_lib/framework/surrogate/ACOSSO.py
+++ b/foqus_lib/framework/surrogate/ACOSSO.py
@@ -1,8 +1,8 @@
 """ #FOQUS_SURROGATE_PLUGIN ACOSSO.py
 
-Surrogate plugins need to have FOQUS_SURROGATE_PLUGIN in the first
-150 characters of text.  They also need to hav a .py extention and
-inherit the surrogate class.
+Surrogate plugins need to have FOQUS_SURROGATE_PLUGIN in the first 150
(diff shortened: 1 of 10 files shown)


foqus_lib/framework/surrogate/ACOSSO.py
```diff
@@ -1,8 +1,8 @@
 """ #FOQUS_SURROGATE_PLUGIN ACOSSO.py
 
-Surrogate plugins need to have FOQUS_SURROGATE_PLUGIN in the first
-150 characters of text.  They also need to hav a .py extention and
-inherit the surrogate class.
+Surrogate plugins need to have FOQUS_SURROGATE_PLUGIN in the first 150
+characters of text.  They also need to hav a .py extention and inherit the
```
perrenyang: I'd like to take this chance to correct "extention" to "extension"

perrenyang:
Awesome, looks like it works (although now AppVeyor is running into some different exception while attempting to test...)

eslickj:
I fixed the spelling error too, which due to copy and paste appeared in a lot of places.

eslickj:
The exception stopping appveyor doen't stop FOQUS from running, but I'll try to track it down.

Issue #408, opened 2019-02-22
Basic Data tab


sotorrio1:
Is there anyone using the basic data tab at all? I was wondering if we should delete this tab completely. As you can see the GUI is all messed up and I was planning on fixing it but only if the module is relevant. As of now, basic data is basically SolventFit and there are a few group boxes and frames that are not even being used.

fsoepyan:
I'm not using the Basic Data tab. The only time I used it was to test it.

eslickj:
I can say I've never used it, and I'm not opposed to removing it.

CSRussell2319:
Not using it. I only went through it to follow examples/test to find bugs. I don't think we officially have anyone supporting the feature either.

sotorrio1:
That's what I was suspecting. We can discuss further during tomorrow's call, but I propose we get rid of the whole module.

ksbeattie:
I'll bring this up to Mike & Ben.

ksbeattie:
I've sent a note to Mike, Ben and John Shinn about this.  I'll let you all know what I hear.

Is it worth keeping it as an "entry-level" module for introducing new users to the tool?  That and maybe a place where simple "smoke tests" live?

Not necessarily trying to argue against removing it, just trying to think if there might still be some value in keeping it that is worth the effort to keep maintaining it.

sotorrio1:
I didn't write the code so I'm not completely sure what this module does and if it's worth keeping it. Let me know what they say and we'll take it from there :)

ksbeattie:
Mike's response is:

> From the below it seems to me that we should allow the surrogates capabilities to generalize the approaches currently in the basic data tab, albeit in a less specific way. I'm ok with the idea of deleting Basic Data in favor of surrogates, with two considerations: 1. That no stakeholder is doing critical work with basic data (I would imagine they aren't, but invite the team to weight in) 2. We consider preserving any residual capabilities in basic data that are "mathematically general" and not available in surrogates.

@sotorrio1 suggested changing a setting to make the tab disappear and will follow up with @ngbrenda on making we are not abandoning any functionality.

sotorrio1:
PR #474

ksbeattie:
Although PR #474 has been merged, removing the basic data tab, we'll close #176 and #207 but leave leave this issue open to revisit on whether to fix those issues for good, or remove the tab for good.

ksbeattie:
[...]

PR #474, opened 2019-04-19
hiding Basic Data tab

sotorrio1:
hiding the basic data tab for now. I suggest after approving this PR, we close issues #176, #207, and #408

### DIFF
--- a/foqus_lib/gui/main/mainWindow.py
+++ b/foqus_lib/gui/main/mainWindow.py
@@ -312,14 +312,14 @@ class mainWindow(QMainWindow):
         #separator
         self.toolbarMain.addSeparator()
         #Basic data action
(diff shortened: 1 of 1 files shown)


fsoepyan:
- I managed to install and open this PR's version of FOQUS.
- The Basic Data button in the toolbar is gone.
- I have no problem with this PR being merged.

sotorrio1:
@fsoepyan thanks for testing this :)

ksbeattie:
@sotorrio1, is it possible to toggle this back on by the user?  As in without editing the code?

sotorrio1:
@ksbeattie no, right now the only way is modifying the flag within the code, but I guess it's something doable...

Issue #456, opened 2019-03-20
Flowsheet data table functions


eslickj:
It seems that delete rows doesn't work on the flowsheet data table.  Need to check all the data table functions and fix if necessary.

ksbeattie:
@eslickj says the delete functionality was fixed in #458 but more testing is needed.

fsoepyan:
- I tested the "Export" in the Menu for the Flowsheet data table
- I managed to generate the CSV file with the results
- I also managed to copy and paste the results into Excel
- The "Export" feature works!

fsoepyan:
For the "Import" in the Menu for the Flowsheet data table:
- I managed to copy and paste the information from the the same CSV file that I generated when I tested the "Export" feature into the data table when I choose to "Import from CSV file"
- Then, I tried to copy and paste by copying a row from the CSV file with the numbers and other information (so that the number of items match the FOQUS data table)
- I got this error message when I choose "Paste data from clipboard":

![image](https://user-images.githubusercontent.com/30811140/56983623-1082ab00-6b52-11e9-968d-8e41e4a10ac9.png)

fsoepyan:
- I selected a row in the Flowsheet data table
- I selected "Row to Flowsheet" under "Edit"
- I got this error message:

![image](https://user-images.githubusercontent.com/30811140/56984920-30679e00-6b55-11e9-87e4-106b639dfcd3.png)

fsoepyan:
Under the "Edit" option:
- In the past, "Clear All Data" works
- I tested "Delete Rows", and this works
- I selected "Add Empty Result", and I got this error message:

![image](https://user-images.githubusercontent.com/30811140/56985090-8dfbea80-6b55-11e9-9fc7-de7ae746046f.png)

fsoepyan:
Also under the "Edit" option, I tried "Edit Set for Selected Rows", I entered a new set name, and I got this error message:

![image](https://user-images.githubusercontent.com/30811140/56985167-b552b780-6b55-11e9-8406-7203c0c98f09.png)

fsoepyan:
The "Calculated Columns" works fine.

fsoepyan:
All items under the "View" option work fine.

fsoepyan:
I selected a row, I selected "Run Selected Rows", and I got this error message:

![image](https://user-images.githubusercontent.com/30811140/56985527-8ee14c00-6b56-11e9-8987-6a52d97723a0.png)

[...]

PR #494, opened 2019-05-22
Fix data table menu

eslickj:
Fixes issue #456 

I removed the functions to run rows and fixed the rest.

### DIFF
--- a/foqus_lib/framework/sampleResults/results.py
+++ b/foqus_lib/framework/sampleResults/results.py
@@ -159,9 +159,15 @@ class Results(pd.DataFrame):
                 # this would happen for cols with less than two .'s
                 continue
             if typ=="input":
-                fs.nodes[node].inVars[var].value = self.loc[row, col]
+                try:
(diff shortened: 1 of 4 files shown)


fsoepyan:
The "Import from CSV file" feature works.

fsoepyan:
Both "Export" features work.

fsoepyan:
The "Row to Flowsheet" feature:

For example, if I select Row 98, and select "Row to Flowsheet", Row 97 ends up becoming the Flowsheet input (instead of Row 98).

This is a minor error, and can be fixed for the next release.

fsoepyan:
"Add Empty Result" works.

fsoepyan:
Everything under the "View" menu works.

fsoepyan:
Given that we only need a minor correction (please see my comments above), I will approve and merge this PR.

fsoepyan:
When I tried to save the file, I got this error message:

image

After closing the file, when I try to open it again, I got this error message:

image

PR #497, opened 2019-05-23
Remove time index from variable browser

eslickj:
Also convert column list for calculated columns to drag and drop


fsoepyan:
Variable explorer works, and the time index is no longer there.

fsoepyan:
The "Calculated Column" list does have a drag and drop feature, but where can we drop the item to?

eslickj:
@fsoepyan you should be able to drop the column into the box where you enter the calculation expression.

eslickj:
Not sure what happened, but this PR seems to be incomplete. I'm just going to tack it onto something else I'm fixing and close this.

PR #502, opened 2019-05-28
Fixing link for copyright.rst file

ksbeattie:
This is just fixing the link for the copyright.rst file.  Hopefully this will still work on Windows...

### DIFF
--- a/docs/source/copyright.rst
+++ b/docs/source/copyright.rst
@@ -1,4 +1,1 @@
-Copyright
-=========
-
 ../../LICENSE.md


eslickj:
Yeah, symlinks don't seem to work in Windows, but now that I know that I can be careful not to touch them.  The other minor issue is that if people try to build the docs in Windows, they'll get some really confusing warning and wen they investigate they they'll be pretty confused by the copyright rst file.  Maybe for the future we should think more about this, but for now looks good.

Issue #519, opened 2019-06-13
SinterConfigGUI Image Update


anujad95:
The GUI Image: _SinterConfigGUI Simulation Meta-Data with Data Completed_ in the Aspen Plus SimSinter Tutorial manual needs to be updated.

Old image:
![image](https://user-images.githubusercontent.com/47067166/59439137-b452ae00-8dc2-11e9-957a-fe601c3e5ed1.png)

Current GUI Image:
![image](https://user-images.githubusercontent.com/47067166/59439228-dd733e80-8dc2-11e9-9ee8-d0872c000ea2.png)

fsoepyan:
PR #520 has been merged, so I will close this issue.

PR #520, opened 2019-06-15
Issue #519 correction

anujad95:
SinterConfigGUI Image Updated in Tutorial Manual
This addresses #519


fsoepyan:
- Is this SimSinter file going to be used for a later tutorial in the manual?
- If so, it's probably best if the metadata (i.e., the title and description) are the same as the previous image.
- But changing the author's name and date should be fine.

anujad95:
I believe that this SimSinter file is used just for the AP Flash example. It doesn't seem to be used in any other tutorial. Would you recommend keeping the title and description same as before?

fsoepyan:
If this is not used in any tutorial, then your image is fine.

PR #524, opened 2019-06-24
Updated SDOE 

sotorrio1:
- Different wording in the GUI
- Improved a few capabilities during the design of experiments
- Added new capability from UQ to add new ensembles using psuade 
- Improved plotting capability with historical data being plotted in the same plot as the design created by the user


sotorrio1:
@candcook or @atowfiq would you mind reviewing this?

Issue #521, opened 2019-06-21
lack of online document interface.


TheHuangs:
It's hard for the user to find the website of the online document. I recommend adding this https://foqus.readthedocs.io/en/latest/ website on the first page of this repository.

fsoepyan:
- If you go to the FOQUS GitHub page (https://github.com/ccsi-toolset/foqus), the "README" has a section called "Documentation and User's Manual".
- The link to the documentation can be found by clicking the word "here" in the second line ("The documentation for FOQUS (including the user's manual) can be accessed by clicking here.").

ksbeattie:
@TheHuangs, thanks for input.

The very first link on the readme (the [![Documentation Status](https://readthedocs.org/projects/foqus/badge/?version=latest)](https://foqus.readthedocs.io/en/latest/?badge=latest)) points to the documentation link you mention.  But I agree that may not be entirely clear to new users.

So, I've open a PR (#525) with some changes that hopefully make it clearer where the documentation lives.

ksbeattie:
Hopefully with my changes in #525, this is better.

PR #525, opened 2019-06-26
Update README.md

ksbeattie:
Addresses #521

Re-organizing and re-wording readme wrt documentation and videos in a hopefully clearer way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,17 @@ Package includes: FOQUS GUI, Optimization Engine, Turbine Client. *Requires acce
 ## Getting Started
 
 ### Install
-See the [installation](https://foqus.readthedocs.io/en/latest/chapt_install/index.html) instructions for details.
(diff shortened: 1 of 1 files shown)


fsoepyan:
- The PR looks fine.
- However, the instruction for generating the html version of the documentation from Anaconda has been deleted.
- @eslickj and @anujad95: Is this OK with you? If you're fine with it, I'm fine with it.

anujad95:
The README file seems to be good. Its fairly detailed. Adding instructions for generating the html version of the documentation would be helpful.

ksbeattie:
Yeah, I removed the html generation bits because I felt that they aren't really needed in the top-level readme now that we have them automatically generated on readthedocs and those instructions do already exist the full documentation (for those who are interested in doing it)

ksbeattie:
btw, I also switched the readthedocs links to point to the `stable` link (which will be the most recent release) rather than the `latest` link (which is the most recent build of master).  My thinking there is that we want people working from releases, unless they really want the latest development work in which case we can assume they know to look the latest link.

PR #543, opened 2019-07-29
Heat Integration Tutorial

anujad95:
This is a simple tutorial that explains heat integration implementation in FOQUS.


fsoepyan:
When I tried to open the "Methanol" SimSinter file, I got this error message:

image

But this could be because I am using Aspen Plus v8.4 instead of v10.

But I managed to successfully upload the SimSinter json file into Turbine and FOQUS.

fsoepyan:
However, when I tried to specify the heat integration plugin, I got this error message:

image

Can anyone please let me know what is wrong?

fsoepyan:
@anujad95: Can you please add an image to show how a correct heat integration tag looks like in the "methanol_HI" node? I think this would be very helpful for the user following the tutorial.

anujad95:
The error which you are getting is obtained with the old plugin. I have fixed the heat integration code in the previous PR. Did you run this with the updated FOQUS files?
In step 5 of the tutorial i have explained how to get the correct list of tags. I shall add an image to support it too.

fsoepyan:
- I downloaded the latest version of FOQUS. The heat integration plugin now works.
- However, I can't run the Aspen file in FOQUS. I believe this is because I have Aspen v8.4 in my computer, but the example file was developed in Aspen v10.

anujad95:
You may want to upgrade Aspen to v10, since the example file has been developed in v10. I shall add a note about this in the documentation too.

PR #564, opened 2019-09-06
UQ ensemble information frame modified

anujad95:
The UQ Ensemble information was not completely seen in the GUI earlier. Hence i modified the frame in qt designer.


sotorrio1:
I don't understand what was the issue here. Is there an issue related to this PR? When I open UQ I have no issues on my screen. I agree with @eslickj that things will work differently in different resolution screens.

anujad95:
The issue was that the UQ ensemble information was not visible properly on my screen, so i enlarged the box size in the GUI. I agree with John on the resolution issue. If the screen is high resolution, viewing the GUI details becomes difficult since they are either hidden, or overlap with each other.

Issue #482, opened 2019-05-08
Running foqusPSUADEClient.py raises issues with python version


sotorrio1:
When running OUU, FOQUS calls the foqusPSUADEClient.py script that takes two arguments; an input file and an output file. 

When running on Windows the user needs to make sure the register keys are in accordance with the documentation updated through @fsoepyan PR #480 

When running on iOS there is no such thing as register keys. I've found that when running a bash script iOS is ignoring the conda environment you're running on; but if we add the following line of code to the beginning of the script, it works:

#!/anaconda3/envs/foqus/bin/python

The problem here is that the user needs to name the environment 'foqus' so this line of code works. If the user decides not to follow the installation instructions and name the environment slightly different, it will fail. 

This issue is more like asking for help/ideas on how to handle this in a generic way.

ksbeattie:
A few ideas to investigate:
- Change the #! line to `#!/usr/bin/env python` which will use the python found on the current's envronment's $PATH
- Make sure PATH has the correct bin on it.

sotorrio1:
Just double checked and the line `#!/usr/bin/env python` works. I'll create a PR to update this

sotorrio1:
PR #490 should take care of this. Once the PR is approved and merged I will close this issue.

eslickj:
On the windows side, I have a suggestion contained in PR #493

ksbeattie:
@sotorrio1 will work on a similar change, as @eslickj used above, for OUU without changes to the registry for the next release.

ksbeattie:
@sotorrio1 will get to this soon.

sotorrio1:
PR #575 should take care of this

fsoepyan:
PR #575 has been merged, so I will close this issue.

PR #575, opened 2019-09-16
Issue#482

sotorrio1:
This should fix issue #482


sotorrio1:
the chmod was being used before so I assume there was an issue with permissions, that's why I left it there. I could test it without it...

sotorrio1:
Just tested and the chmod is not needed at all. Just removed it.

eslickj:
If @fsoepyan, @anujad95, or @jmorgan29 can confirm it works and approve, it still looks good to me.

PR #576, opened 2019-09-17
Modifying Optimization Solver Descriptions

anujad95:


anujad95:
PS: I have made box size changes to optSetupFrame_UI.ui, however this is creating a conflict with the main repository master branch. What can be done in this case @eslickj @ksbeattie ?

Issue #511, opened 2019-05-31
Tool Validation with more examples


anujad95:
According to the documentation, the Foqus tool has been implemented for a few cases which include simple node simulations, flowsheet with linked simulations (BFB & Cost optimization setup), recycle, etc. As the tool is being developed further with new, improved features, it might be worth to validate it simultaneously with more complex examples, & different types of carbon capture systems like solvent based (MEA) & membrane based ones. We can use the process models developed within the ProcessModels_bundle of the CCSI toolset. After implementing these, supported with concrete problem statements, we can include some of them as part of the tutorial manual/validation section. This would help in making the tool more convincing to the user, and a more extensive demonstration of its capability. Please let me know your thoughts/suggestions on this.

ksbeattie:
@jmorgan29 will update the MEA_ssm model, see [MEA_ssm #6](https://github.com/CCSI-Toolset/MEA_ssm/issues/6) then he and @anujad95 will work on adding an example and documentation to FOQUS

fsoepyan:
I think the example problems in the user's manual is fine, as the purpose of these examples is to get the user familiar with FOQUS.

But to motivate the user to use FOQUS, I think including the list of publications where FOQUS has been used will be a good first step.

So far, I have compiled this list of publications:

Chen, Y., Eslick, J.C., Grossmann, I.E., Miller, D.C., 2015. Simultaneous process optimization and heat integration based on rigorous process simulations. Computers and Chemical Engineering 81, 180–199.

Gao, Q., Miller, D.C., 2015. Optimization of amine-based solid sorbent chemistry for post-combustion carbon capture. Paper presented at: 2015 International Pittsburgh Coal Conference; 5–8 October 2015; Pittsburgh, PA, USA.

Ma, J., Mahapatra, P., Zitney, S.E., Biegler, L.T., Miller, D.C., 2016. D-RM Builder: A software tool for generating fast and accurate nonlinear dynamic reduced models from high-fidelity models. Computers and Chemical Engineering 94, 60–74.

[...]

PR #577, opened 2019-09-18
List of Publications

anujad95:
Adding list of CCSI2-FOQUS related publications


anujad95:
Partly solves issue #511

PR #598, opened 2019-10-11
Foqus Cloud Update:   Refactor foqus_service

boverhof:
Moved code into `foqus.service.flowsheet` module and made the `cloud/aws/foqus_service.py` a WindowService stub that calls that module.  Added very simple pytest.


foqus_lib/unit_test/nodeVarListTest.py
```diff
@@ -1,4 +1,4 @@
-from foqus_lib.framework.graph.nodeVars import *
+from foqus_lib.framework.graph.nodeVars import nodeVarList
```
ksbeattie: I think that should be left as `*` for now.

I've asked @eslickj to rename the files under the `foqus_lib/unit_test/` dir (from `*Test.py` to `*_test.py` so pytest will find and run them (#607) but they need some love to get them running properly.

Issue #535, opened 2019-07-15
ALAMO Stop button not working well


anujad95:
In case we want to terminate the ALAMO modeling within FOQUS, it takes a very long time for the iterations to stop. Is there a way to improve this?

ksbeattie:
This lag in time happens because the model is not going to notice that the stop button has been pressed until the currently executing model completes.  @anujad95, @eslickj and @boverhof will explore ideas on how to improve this.

-

eslickj:
We had a few ideas on this.
1. Document current behavior better
2. As a work around and to better pin down what needs done, use task manager to kill Aspen and document what processes need to be killed.
3. Write a small script to kill Aspen (and other).
4. Try to have FOQUS send a rest kill command to consumer (should usually work but not %100)
5. Integrate method to directly kill Aspen (or similar).

anujad95:
I have tried to use the methods killTurbineJobs() and stopAllConsumers() separately, followed by process.kill(), when the user presses the stop button. However, this does not seem to work. Any suggestions?

ksbeattie:
@anujad95 will work with @eslickj on this.

anujad95:
I had a discussion with @eslickj regarding this. It seems that the "stop" button within the surrogate tab is not hooked up properly within the base code. I shall look into this. In the meanwhile, users can kill aspen, and alamo process directly from the task manager to stop surrogate building. This is working fine.

anujad95:
I made changes to the gui base code surrogateFrame.py, to connect the ALAMO stop button properly for terminating the runs. I tested it with an adaptive sampling example that takes a while to converge, and tried to press the stop button during the runs. Although it takes some time to actually stop the surrogate building, a user termination message is displayed in FOQUS.

PR #626, opened 2019-11-11
ALAMO stop button fixed

anujad95:
I made changes to the gui base code surrogateFrame.py, to connect the ALAMO stop button properly for terminating the runs.


anujad95:
This should fix issue #535

Issue #457, opened 2019-03-20
Add instuctions for data filter entry


eslickj:
It's almost impossible to figure out how to write a data filter without looking at the documentation.  I think we may want to add a frame to the dialog box that provides some basic instructions.

anujad95:
Data Filters are present in the Flowsheet, Uncertainty, and Surrogates tab. Just to confirm, we should add data filter instructions to this dialog box in each of the modules?:
![image](https://user-images.githubusercontent.com/47067166/65781902-89def380-e11a-11e9-9135-885c7315ec16.png)

PR #628, opened 2019-11-12
Adding data filter instructions

anujad95:
This should take care of issue # 457. However, sizing the message box properly is an issue.


anujad95:
#457

Issue #631, opened 2019-11-14
Syntax Error in nodePanel.py


anujad95:
An import code line in nodePanel.py has incorrect syntax:
from foqus_lib.gui.pysyntax_hl.pysyntax_hl import *
This leads to a code error in subsequent code lines as well. 
class pysyntax_hl does not exist, and its PythonHighlighter class instead. Moreover, the syntax for importing this class from the file should be from foqus_lib.gui.pysyntax_hl import PythonHighlighter. This change would be made in the corresponding PR.

anujad95:
PS: There may be issues with the suggested change. Please give suggestions on the correction required.

ksbeattie:
@anujad95, what's the priority on this?  Is this something that should be part of the Dec release?

anujad95:
I observed that it does not affect anything directly in FOQUS, however, the reason why code error is shown in the python file, needs to be checked.

anujad95:
As seen in the code screenshot below, line 17 is showing an error:
![image](https://user-images.githubusercontent.com/47067166/69736446-85c25880-1100-11ea-9cf6-f10ac1c740c6.png)

anujad95:
It seems that this error is specific to the spyder environment, and hence doesn't require any changes in the base code.

ksbeattie:
Can you see what tool Spyder is using to do this code analysis?  There are a few out there it might use: pylint, pep8, etc.  Knowing which one it uses might help better understand the error it is reporting.

Or in the meantime, can you replace the '*' with the specific objects needed from that module?  Then see if the error goes away?

PR #632, opened 2019-11-14
Syntax correction in nodePanel.py

anujad95:
Should take care of #631


anujad95:
PS: I am unable to open FOQUS with this change. Please suggest the required corrections.

Issue #593, opened 2019-10-09
Additional Optimization Plot


anujad95:
As an enhancement, an additional plot of objective value v/s iteration can be included in FOQUS, that displays the actual objective value at iterations after regular time intervals, instead of plotting just the "best so far" objective values. This would help the user understand the actual progress and behavior of the optimization solver.

anujad95:
This will also be followed by displaying the iteration number corresponding to the optimal solution, in the message window, after the solver converges, so that the user can access it accordingly from the flowsheet result table. 
This issue is a modification of #579

ksbeattie:
fixed in #644

PR #644, opened 2019-11-25
Optimization Plot Change

anujad95:
This should take care of #593 
I have also attached a simple optimization foqus file, in which the plot changes can be viewed by running the optimization.

### DIFF
--- a/foqus_lib/framework/optimizer/NLopt.py
+++ b/foqus_lib/framework/optimizer/NLopt.py
@@ -249,10 +249,10 @@ class opt(optimization):
             self.updateGraph = True #this flag is for GUI if true the
                                     #flowsheet display needs updated
             self.resQueue.put(["BEST", [self.bestSoFar], x])
-        else:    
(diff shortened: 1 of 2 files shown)


anujad95:
I have made changes only to NLOpt now. Once this is approved, i will make changes to other solvers.

anujad95:
The changes that i have made basically create 2 optimization subplots instead of one. The first plot with red markers would depict the actual solver progress (objective function values after each function evaluation). The second plot with blue markers would depict the best values of objective function evaluated over the iterations. 
At the end of the optimization, a message would be displayed, with the iteration corresponding to the minimum objective function value.

anujad95:
OptNewPlot

eslickj:
This looks pretty good, but looking at the top plot, I wonder if we shouldn't use a log scale.

anujad95:
I'm not sure if the top optimization plot has log scale. These are the actual objective function values.

Issue #607, opened 2019-10-25
Fix unit_test test


ksbeattie:
After PR #606, @eslickj, could you now get these to work with pytest?  They currently aren't getting run by pytest because the .py files aren't named correctly.  If you rename them to be `_test.py` rather than `Test.py` they should then get picked up and run.  But, of course the tests are failing, which is beyond my understanding to properly fix them.

anujad95:
Is this issue resolved?

eslickj:
Not that I know of.  At least I didn't do anything to resolve it.

ksbeattie:
The files I'm talking about here are in `foqus_lib/unit_test/`

ksbeattie:
Closed by #647

PR #647, opened 2019-11-28
Fixing Unit Tests

anujad95:
This is my first attempt to fix issue #607. I changed the names of the test files as required, ran pytest, and went through the errors. I have made some changes in the unit tests, as well as nodeVars.py (trying to add missing code lines and syntax corrections). 
**nodeVarList_test**:
- I think this is failing due to lack of vector variable support, so i have completely commented vector variables, and related tests in this file. 

**nodeVars_test**:
- I have removed tests related to vector variables, time steps, and set shape
- commented lines 170, 172
Further changes in the code will be made as suggested, and I will remove all commented code lines, once the testing requirements are finalized.


eslickj:
Just had one comment that may need addressed, and remove the commented lines.  Other than that seems good.

PR #437, opened 2019-03-08
Fix ACOSSO and BSS-ANOVA

eslickj:
When we updated the results format to Pandas we lost the exportVarsCSV function.  I guess since it has been over a year sicne that change was made it's safe to say ACOSSO and BSS-ANOVA aren't used much.

Fix for issue #433 and #434.

Issue #34, opened 2017-10-02
Surrogates: ACOSSO check for RScript not RGUI and improve stop button


fsoepyan:
From JIRA:
- Reporter: Natalia Kitch
- Assignee: John Eslick (@eslickj)
- Original title: ACOSSO check for RScript not RGUI and improve stop button
- Created: 04/Nov/15 2:55 PM
- Updated: 13/Jan/16 8:44 AM

Description:

How long ACOSSO tutorial has to run, as it is running very long, and it looks like doing nothing, so that it needs to be killed, as Stop function/button doesn't stop it:

Starting ACOSSO

Model File:     acosso_fit.rds
Py File (UQ):   ACOSSO_Tutorial_UQ.py
RScript Path:   C:\Users\kitch2\Documents\R\R-3.2.1\bin\x64\Rgui.exe
Sub-directory:  acosso
X data file:    acosso\xdat.csv
Y data file:    acosso\ydat.csv
Order:          2
CV:             bic
------------------------------------
Exporting Data...
    Inputs: ["eq.x1", "eq.x2"]
    Output: ["eq.z1", "eq.z2"]
Runinng ACOSSO...

Chat history:

Natalia Kitch (Inactive) added a comment - 05/Nov/15 10:27 AM 

It appears that "Rgui.exe" was set instead of "RScript.exe" in RScript Path, which caused the hanging.
It would be nice if FOQUS could give a warning in this case, and Stop button would stop the run.
Perhaps, the issue can be closed.

John Eslick added a comment - 16/Nov/15 7:21 AM 

Will add checks in January release. Also will try to improve the stop button. The stop button currently stops simulations running for surrogate methods, but doesn't stop the surrogate method itself.

ngbrenda:
FOQUS needs to prompt user to explicitly set their R path, the same way that PSUADEPATH is set.

Perhaps we need a quick R test?

anujad95:
We may want to know the actual number of users currently working with ACOSSO.

ksbeattie:
@jmorgan29 says he will take a look at this and see if this is still an issue

jmorgan29:
There still appear to be some issues with using ACOSSO in FOQUS:

[...]

PR #656, opened 2019-12-13
Syntax changes to acosso and bss anova driver files

anujad95:
Some basic syntax changes, which might take care of #437


anujad95:
#34

Issue #682, opened 2020-02-11
Upgrade to new pandas API


ksbeattie:
On Jan 29 pandas released their 1.0 with a new non-entirely backward compatible API.

This was first reported in #679 and temporarily addressed in #681

We should update our use of pandas the new API following their [Backwards incompatible API changes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.0.html#backwards-incompatible-api-changes) doc.

fsoepyan:
- @ksbeattie: Thank you, your PR works.
- I managed to create a new environment in Anaconda called foqus_test
- I managed to download and install the latest version of FOQUS successfully
- I managed to open FOQUS, and open a file in FOQUS successfully

anujad95:
This works for me too

ksbeattie:
So, this issue is for us to fix our usage of pandas so that it works with their latest release.

Then we change from our requierments.txt the `pandas==0.25.3` fixed version line (back to just `pandas`) so that we are not stuck using an old version of pandas (which will eventually bite us).

fsoepyan:
I will close this issue, because I believe this issue has been resolved with #686.

PR #686, opened 2020-03-03
Fix pandas for compatablity with 1.0

eslickj:
I fixed the incompatibility with new Pandas 1.0.


ksbeattie:
Waiting for @fsoepyan to reproduce his steps in #682 to verify this is fixed, before merging.

Issue #660, opened 2019-12-18
FOQUS pypi package too large


ksbeattie:
I thought I'd try to see if I could create a pypi package for FOQUS.  It seems I can, but I can't then upload it because it is over their 60MB is size limit (it's 78MB).

I could ask for an increase (they'll want an explanation), but before that I thought I see if there is a way that we could get under the 60MB file size limit.  

I know that there are several duplicate large files found in #168 but I gave up on trying to address that.  Any thoughts on if we could re-open that idea?  I think we could then get under the 60MB size limit.

anujad95:
I shall look into the duplicate files and we can try to get it under the 60MB size limit.

eslickj:
I think in the long run the solution is to move examples out to a separate repo.  We have some big example files and installing them with pip may not make a lot of sense.  I think we had a similar problem in IDAES (although not size related) which led to the creation of an examples repo.

ksbeattie:
Closing this now that PR #700 is merged, though I'm sure we haven't heard the end of this problem.  Ideally moving the example out into their own repo will be the final fix.

PR #688, opened 2020-03-09
[WIP] Clean up duplicate or unused files

ksbeattie:
Addresses #660, by removing duplicate files or re-organizing the tutorial so that duplicate files do not need to live in the repo (and a pypi package).


ksbeattie:
@fsoepyan, I couldn't find these files referenced anywhere.  Are they safe to remove?

    examples/test_files/Heat_Integration/Model_Files/BFB_System_ACM/BFB_2ads_2rgn_old.acmf
    examples/test_files/Heat_Integration/Model_Files/BFB_System_ACM/BFB_2ads_2rgn_old.json

ksbeattie:
Ok, thanks.  I've removed the following dirs:

    examples/tutorial_files/Flowsheets/Tutorial_2/
    examples/tutorial_files/Optimization/Tutorial_1/
    examples/tutorial_files/SimSinter/Tutorial_1/

As all the files in each, are duplicated in:

    examples/test_files/Optimization/Model_Files/

or

    examples/test_files/Optimization/

Then I updated the docs so that that dir is referenced.

_**Could people please review the tutorials so they still make sense and work correctly?**_

Removing these duplicate files, even though I realize this crosses the test_file/tutorial_files boundary (which will have to be completely re-done if/when we move all the example/tutorial_files/ into it's own repo) *should* (I hope) make it so that we can now create a pypi package.

fsoepyan:
I looked through the changes, and I think they're fine.

PR #698, opened 2020-03-18
Build docs and check for errors

ksbeattie:
Get CircleCI to build docs and fail if there are any errors (or even warnings).


ksbeattie:
This PR won't pass circleci tests until PR #697 is merged.  Which is exactly the point, I'm thinking  we don't want circleci tests to pass if there are problems with generating the docs.

PR #700, opened 2020-03-25
pip install ccsi-foqus

ksbeattie:
Changes so that a pypi package can be created.

There's a few changes here to where requirements live and how to setup a developers env.

Regular "end-user" requirements now live in `setup.py` in the `install_requires` list.  They should **_not_** live in `requirements.txt` anymore.  All that's left in `requirements.txt` now (beyond comments) is the line `-e .` which means "look in setup.py".

I've also moved "developer" requirements into a new `requirements-dev.txt` file so that people who don't want those can not bother with them.

This changes (a little) how a developer setup up their environment.  The new instructions are (I will put these in the docs, but here they are for now):

1. Create a new conda env (here named "foqus")
    ```
    conda deactivate
    conda remove --name foqus --all
    conda create --name foqus python-3.7
    conda activate foqus
    ```
2. Install requirements:
    ```
    cd <your foqus working dir>
    pip install -r requirements.txt  # or pip install -r requirements-dev.txt
    ./foqus.py   # note the "./" there as the above line will also install a foqus.py in your conda env
    ```
There's no longer the need to run `python setup.py [install|develop]`.  (In fact, you don't want to it doesn't install PyQt5 correctly.)

ToDos:
 - [ ] Update developer docs for how to create developer env
 - [x] Add a wiki page on how to build package for pypi
 - [ ] Update install docs on how to use `pip install ccsi-foqus`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
-            pip install -qr requirements-dev.txt
+            pip install --progress-bar=off -r requirements-dev.txt
 
       - persist_to_workspace:
           root: ~/repo
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,12 +44,6 @@ jobs:
             pip install -q pytest coverage pylint
 
       - run:
-          name: Run PyLint
-          command: |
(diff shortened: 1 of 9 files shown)


ksbeattie:
Related to #660

Issue #701, opened 2020-03-26
Update docs on pip install


ksbeattie:
As described in PR #700, documentation on how to install FOQUS needs to be updated to describe how use either the PyPi package or the Windows installer.

Also, some kind of documentation should be created for developers on how to setup their development environment.

- [x] Install docs
- [x] developer docs

PR #702, opened 2020-03-30
[WIP] Improve `pip install` on Windows

ksbeattie:
When installing via `pip install ccsi-foqus` on Windows, the batch files do not get created properly.  Specifically, the code generating them has several assumptions which may or may not be true: 

- It's run in a conda env (could just be virtualenv, the system python a/o with `pip install --user`
- There's a copy of the source in the current dir (the goal is for the end-user to NOT need the source)
- It is run on Windows (setup.py might be run on unix to generate a wheel that could be installed on Windows)

The larger goal here is to make it easier for a Windows user to install and start FOQUS.

It appears that [bdist_wininst is going away with python 3.8](https://discuss.python.org/t/deprecate-bdist-wininst/1929).  Hopefully there's a way to still work with a [wheel](https://pip.pypa.io/en/latest/user_guide/#installing-from-wheels) that will work on Windows.  Ultiumately a full-blown Windows installer is the way to go, but hopefully there's a mid-point here with pip and wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -154,4 +154,4 @@ Windows:
         - foqus.bat: run FOQUS and close cmd window when FOQUS exits
         - foqus_debug.bat: run FOQUS and leave cmd window open  
 ==============================================================================
-"""
+""")
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,8 @@ dist = setup(
     include_package_data=True,
     scripts = [
         'foqus.py',
+        'foqus.bat',
+        'foqus_debug.bat',
         'cloud/aws/foqus_worker.py',
         'cloud/aws/foqus_service.py',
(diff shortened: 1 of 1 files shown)


ksbeattie:
https://pythonwheels.com/ has some good info on creating wheels.

ksbeattie:
After a `pip install ccsi-foqus` on Windows, in a conda env, this works to start foqus:

    python %CONDA_PREFIX%\Scripts\foqus.py

ksbeattie:
This is related to: #701

ksbeattie:
@ksbeattie will try to get to this.

ksbeattie:
closing this as #752 takes care of nearly all of what I was trying to achieve here.

Issue #648, opened 2019-12-02
Optimization samples in csv format


anujad95:
In the optimization tab, the format in which the sample variable data can be loaded in csv format, in the "Samples" section, is not mentioned. Once this is checked, it can be added in the documentation.

jmorgan29:
The instructions for using this feature are now explained clearly in the documentation, but there still seems to an issue with using it. In an example I am working with, the uploaded samples look like this:
![OptimizationSample](https://user-images.githubusercontent.com/11306727/76241894-19c9b300-620c-11ea-9b31-91f003e00f55.png)

The first column of data is completely populated with NaN, which is inconsistent with the csv file that is being uploaded. Here is a screen shot of the csv file:
![OptimizationSample](https://user-images.githubusercontent.com/11306727/76242019-53022300-620c-11ea-8f26-79bb37149345.png)

ksbeattie:
@fsoepyan will take a look at this and try to give some guidance on where in the code this issue might exist.

@eslickj suspects that this is likely using a pandas dataframe and might get the headers wrong somehow.

fsoepyan:
- I used the test file that @jmorgan29 sent me.
- I saved the CSV file as a "CSV (MS-DOS)" for the file type (referenced in #263).
- I went to the Optimization tab in FOQUS.
- I went to the "Samples" tab.
- I clicked "Generate Samples", and chose "Load From File".
- I managed to get these samples:

![image](https://user-images.githubusercontent.com/30811140/76431184-a3ea5680-6387-11ea-9018-397e84178571.png)

- @jmorgan29 or @anujad95: Can you please check to see if saving the CSV file as "CSV (MS-DOS)" resolves this issue?

anujad95:
I'll take a look at this Byron. I'll try to save the csv file as MS-DOS. Thanks.

ksbeattie:
@anujad95 will look into the csv reading code, to make this less specific to the MS-DOS format.  @ksbeattie will also help.

ksbeattie:
One possibility might be to use https://docs.python.org/3/library/csv.html?highlight=sniffer#csv.Sniffer as is done elsewhere in our code.

ksbeattie:
@anujad95  will update some docs on this, then close it.

ksbeattie:
Closed with #711

PR #711, opened 2020-04-21
Optimization samples csv file format

anujad95:
This should take care of #648


docs/source/chapt_opt/tutorial/parameter_est.rst
```diff
@@ -171,7 +171,7 @@ objective function calculation.
 20. For larger sample sets, **Generate Samples** has an option to load
     from a CSV file. The CSV file must be saved as "CSV (MS-DOS)" as the file type, as follows:
 
-.. figure:: ../figs/optim_samplevars_csv.png
```
ksbeattie, line 174: It looks like this is the only place this image is referenced.  If it's getting replaced by a new file (optim_csv_samples_format.png) seems like the old one should be removed, right?

Issue #720, opened 2020-05-04
File Format Documentation Missing


sotorrio1:
The file format documentation referenced in different sections of the UQ chapter is being missing for a while. It would be good to have it back in the online documentation as a reference for users to know what is the right format in each situation.

ksbeattie:
Is this related to #605?  The `ap:psuadefiles` parts?

sotorrio1:
oh my god, you're right! Totally forgot about that issue. Yeah, with the addition of this section and a few edits I've made, I think I fixed all the broken references you mentioned in that issue

PR #721, opened 2020-05-04
Adding File Formats Documentation

sotorrio1:
Bringing back the file format section from the original FOQUS documentation and fixing some broken references to it in the UQ chapter. After merging this, issue #720 can be closed.


ksbeattie:
@sotorrio1: I'm not 100% sure if that last replacement of the "ap:psuadefiles" is correct.  Please take a look.

PR #724, opened 2020-05-08
SDoE: Creating candidate documentation

sotorrio1:
There was a small space missing that was creating the wrong outcome for an enumerated list in the creating a new candidate section within the SDoE documentation.


sotorrio1:
I've updated the documentation with the latest feedback received from LANL. PR is now ready for review.

Issue #672, opened 2020-01-27
Warning message while opening FOQUS


anujad95:
I am getting a warning message while opening FOQUS due to some internal updates:

C:\Users\deshpana\AppData\Local\Continuum\anaconda3\envs\foqus\lib\site-packages\sklearn\externals\six.py:31: FutureWarning: The module is deprecated in version 0.21 and will be removed in version 0.23 since we've dropped support for Python 2.7. Please rely on the official version of six (https://pypi.org/project/six/).
  "(https://pypi.org/project/six/).", FutureWarning)

ksbeattie:
I chased this down to a problem in mlrose.  I've opened a PR in that project, which I believe will fix it: https://github.com/gkhayes/mlrose/pull/50

ksbeattie:
@sotorrio1 will look into the possibility of replacing mlrose with mlrose-hiive.

PR #728, opened 2020-05-26
Switch to using mlrose-hiive rather than mlrose

ksbeattie:
Addresses #672 by using mlrose-hiive rather than mlrose.


ksbeattie:
@sotorrio1, @ngbrenda, @atowfiq I added a new file with one test for the `mat2tuples` method there.

Anyone want to take a whack at writing a similarly simple test for the rank() method?

If someone could give me examples of some valid input/output I could try...

ksbeattie:
fyi: the order.rank() test uses hard-coded values taken from debug statements added to it while running through the SDoE USD example using a candidate.csv file.

Also:
- I added a default arg to rank() so that I could make the test faster.  This is the max_attempts arg to the mlrose.genetic_alg() call, which still has a default of 100 but for the test I use 5, which makes it take 8 sec rather than 2 min (on my VM).
- The save=True arg does not appear to be used anywhere.  It also appears that if it ever were be passed as true, rank() would return None - is that intended?
- I replaced the print() with a _log.info() call.  Looks like there's more in sdoe.py that could be done that way to be consistent with the rest of the app.
- There really should be some GUI indication that the "Order Design" button has been pushed and it's working.  Otherwise one could easily think the button does nothing at all, unaware it is just working away on things.

PR #752, opened 2020-06-17
Creating a `foqus` an entry_point

ksbeattie:
I believe that if we can move the current top-level foqus.py script into a module and call it using setup.py/distutils "entry_point" method (as described [here](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point)) this will simplify the installation and execution for users.

Particularly for Windows users as I believe a wrapper .exe is generated to call the python function defined in the entry_point.

This PR is a different run at what PR #702 is trying to do and would replace it, if successful.


ksbeattie:
Note that the "new" foqus.py here is just a copy of the existing top-level foqus.py with a `def main` replacing it's `if __name__ == '__main__':` line.  Tho more changes are surely needed.

foqus_lib/foqus.py
```diff
@@ -0,0 +1,760 @@
...
+    A signal handler to cause a siginal to raise a keyboardinterupt exception.
+    Used to override a default signal like SIGINT so the FOQUS consumer process
+    can shutdown cleanly. The FOQUS consumer catches the keyboardinterupt
+    exception as one (slighlty unreliable) way to shut down.  Seems ctrl-c
+    causes keyboard interupt exception and SIGINT signal, hense need to change
+    SIGINT handler.
+    """
+    raise KeyboardInterrupt()
+
+def main():
```
eslickj, line 188: Add ```global dat``` here, since you moved this into a main function.  Then this should work.  I guess we really need to rework foqus.py, but this is the easy fix for now.

PR #755, opened 2020-06-24
Remove old foqus.py and reworked the batch file.

eslickj:
This removes the now redundant (PR #752) top-level foqus.py and updates the setup.py file accordingly.  The batch file generation was also updated.


eslickj:
@ksbeattie, I was hoping the foqus.exe would eliminate the need for foqus.bat, but it looks like to run foqus.exe you need to have some conda related environment variables set, so we still need the bat file if we want a double click to foqus icon to start type thing.

As far as I can tell with conda, the foqus.exe doesn't require you to activate your foqus environment, so it does potentially eliminate one step.

eslickj:
Okay, a test failed because "where foqus.exe" failed in setup.py in the batch file generation code.  It may have only worked for me because I previously installed foqus.  Not sure, but I commented it out for now until we decide what to do with it.

eslickj:
@ksbeattie, I finally got around to trying your shortcut script and as far as I can tell it works great.  I guess we have to add winshell to the requirements.  We need to decide how we want to execute it too.  As part of setup would be good, but there may be some people that don't want us to automatically add a desktop icon.  Could add it to the scripts or make a foqus subcommnand?

The cool thing about this is you can pin it to the taskbar and delete the shortcut and it all still works.  We can probably go ahead and delete the batch file code.

ksbeattie:
Hey @eslickj, I've moved the make shortcut code into foqus.py so it is now just a subcommand.  But it isn't getting the icon correct.  Do you know if there is that icon already under the installed foqus_lib tree after setup.py is run during an install?

PR #756, opened 2020-06-24
Fixing NUSF include column feature 

sotorrio1:
Fixing the include column feature in NUSF (it wasn't working properly before)

@atowfiq Would you mind testing? Keep in mind now we're doing more calculations with pandas data frames and that's slowing things down drastically. I'm working on options to speed things up, but this will happen in the next release. 

@ksbeattie for the pandas speed up I'm trying a package called modin. I've added modin and modin[all] to the setup.py file but it would be great if you can take a look at that. modin is the package and the "pip install modin[all]" command is needed to install the engines.


sotorrio1:
@ksbeattie BTW I ended up removing 'modin' because it wasn't helping at all.

Issue #713, opened 2020-04-24
Autosave FOQUS Sessions


anujad95:
It would be helpful to have this feature, to ensure that if FOQUS shuts down unexpectedly, UQ ensemble simulations, or optimization, surrogate building runs get saved automatically. It will be convenient and time saving for the users.

eslickj:
We have this at least partly implemented in some places like optimization.  I think I had a feature where UQ would be able to pick up where it left off (at least on the cloud).  We need to take a closer look at what's there and how best to complete this.  For UQ it may be best to run in "remote" mode even locally.  That would potentially activate the feature to allow you to pick up where you left off.

ksbeattie:
Doesn't look like this is going to be done anytime soon, so closing for now.

PR #788, opened 2020-08-05
Fix commandline optimzation

### LINKED ISSUE #785: Check Command-line UQ and Optimization Runs
Command line FOQUS runs may be broken.  We need to check them, fix them if necessary and add tests.  This may have some overlap with improvements to allow the FOQUS consumer to run UQ and optimization jobs (issue #769 point 1).

eslickj:
Fixes #785  To use it do something like this:

```sh
foqus -l C:\Users\EslickJ\git\foqus\examples\test_files\Optimization\Opt_Test_01.foqus --run opt -o test_opt.foqus
```

The paths in the command may be a little confusing.  Relative paths in the file to load are relative to where you run the command.  The output is relative to the foqus working directory.  We should be able to take this and turn it into tests pretty easy.  I know we want to write smaller scale tests, but writing some optimizations test would be a simple way to get the test coverage up by a lot really quick.


ksbeattie:
This is good to know, perhaps we could also make a pytest hook into it, so it shows up in our coverage numbers.

eslickj:
@ksbeattie, I guess that's what is intended.  The optimization module auto saves the optimization results.  So basically focus loads the the session you provide with the -l argument.  When FOQUS does the optimization, it automatically saves it when it's done.  Then it saves it to the new output file specified with -o.   I think there may be an option to not auto save to the original file, but I not sure off-hand.

We can potentially sort this out when we take up issue #713.  As it is, it's not exactly a bug, it's just maybe a weird setup.  I think we can merge this.  Then we can put some thought into how we save sessions.

PR #796, opened 2020-08-19
Surrogate based optimization improvements

anujad95:
The following changes have been made related to the SM Optimizer:
1. ipopt installation instruction in the documentation
2. Plugin option to select multistart method for initializing the optimization problem at each iteration


anujad95:
Yes, the IDAES distribution of ipopt should ideally be used by the users. We can provide a reference for it in this documentation.The pip install for ipopt is the easy way out, and i think it uses mumps as the linear solver. It is an option for users who are not using IDAES.

PR #802, opened 2020-08-26
Total node script mode documentation 

anujad95:
The documentation for 'total' mode in node script has been improved.


docs/source/chapt_flowsheet/reference/flowsheet.rst
```diff
@@ -271,24 +271,28 @@ simulation has been run.
...
+runs the script before a node simulation, (2) **Post** runs the script after a node
+simulation, and (3) **Total** scripts how a node runs the script and simulation.
 
-Figure :ref:`fig.post.calc` illustrates the **Node
-Script** tab of the **Node Editor** with calculations for an
-optimization test problem.
+Node scripts can be any valid Python code. The input and output
+variables for node scripts are stored in dictionaries x and f. The
+dictionary keys are the variable names. The x dictionary is used to access and/or
+update the node input variable values before and after the calculations are executed.
```
eslickj, line 281: The x dict is a copy of the input variable values when the node script starts.  The f dict is a copy of the output variables when the node start.  At the end of the node script the f values are copied back to the node. That means that if you want to change an input to a simulation you need to not use x and access the node vars directly.  Also if you do something like run a simulation and the outputs change, f is out-of-date.

The text here creates a different impression.   I know I said a documentation update would be enough here, but now I'm thinking we should change the code to make the text you wrote here true.  Instead of value copies, we could make x and f pointers to the actual variables values, so you can change the originals and they are always up-to-date.

I'm not totally sure how to fix x and f off-hand, but we could just do away with them and add docs about directly accessing the node.  I'll try a couple things and see if I can find an easy way to make f and x work more as expected and get back to you.

ksbeattie:
@eslickj is planning on making the deeper changes to the flowsheet code for this and not just rely on a documentation change.

ksbeattie:
This will be picked up again after the vector support is in place (PR #876)

ksbeattie:
This is now being worked on, and will support both scalars and vector variables.

ksbeattie:
This is now replaced by #912

Issue #600, opened 2019-10-22
pylint -E fixes


ksbeattie:
Try to get pylint -E passing on all our code.

As least with this invocation:

    pylint -E foqus_lib --extension-pkg-whitelist=PyQt5 --ignored modules=win32process,win32api,adodbapi

See the circleci run of this here:

https://circleci.com/gh/CCSI-Toolset/FOQUS/2

ksbeattie:
While working on the files in `foqus_lib/unit_test/`  I'm not sure why pytest didn't pick up these tests.  It appears they are not properly named:
```
12:44:46	 From Joshua Boverhof : def test_results_empty():
12:49:14	 From Joshua Boverhof : “pytest will automatically collect unittest.TestCase subclasses and their test methods in test_*.py or *_test.py files.”
12:51:55	 From Joshua Boverhof : [pytest]
python_files = test_*.py check_*.py example_*.py
12:53:19	 From Joshua Boverhof : JFYI: http://doc.pytest.org/en/latest/reference.html#configuration-options
```
I'll try some renames to get them tested and fix the pylint -E errors.

PR #804, opened 2020-09-15
Pylint-specific configuration and directives to address false-positive errors

lbianchi-lbl:
This PR implements changes aimed at getting rid of false-positive errors, i.e. portions of the codebase that are erroneously interpreted as errors by the pylint static analysis.

This is part of the efforts towards addressing #600. Specifically, this PR contains only changes that affect pylint exclusively, and thus **do not alter** the runtime behavior of the code in any way, i.e.:

- pylint configuration
- pylint commands in the CI configuration
- pylint directives expressed as comments throughout the codebase

The idea is to keep those sets of changes that do affect the code itself (and that therefore could alter the runtime behavior) in dedicated PRs, to streamline both the fixes and the review process. In other words, reviews for this PR can concentrate entirely on the pylint analysis, without having to take into account compatibility, introducing regressions, or other code-specific considerations.

In addition, these changes aggregate most of the configuration into the pylint RC file located at `.pylint/pylintrc`, allowing to share the same settings among different use cases in addition to CI checks, e.g. direct invocation, and integration with IDEs/editors supporting static analysis with pylint.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
           name: PyLint
           command: |
             . venv/bin/activate
-            pylint -E foqus_lib --extension-pkg-whitelist=PyQt5 --ignored-modules=win32process,win32api,adodbapi,scipy.spatial || true
+            pylint --rcfile=.pylint/pylintrc --disable=W,C,R,I foqus_lib || true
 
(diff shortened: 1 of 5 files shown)


ksbeattie (approved):
LGTM

Issue #805, opened 2020-09-15
pylint errors in foqus_lib/framework/optimizer/SM_Optimizer.py


lbianchi-lbl:
Parts of the efforts towards #600.

- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L364` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L364)) Undefined variable 'Var'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L369` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L369)) Undefined variable 'Var'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L395` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L395)) Undefined variable 'Objective'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L397` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L397)) Undefined variable 'ConstraintList'
- [ ]  `foqus_lib.framework.optimizer.SM_Optimizer:L428` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L428)) Undefined variable 'Warmstart'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L469` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L469)) Undefined variable 'value'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L499` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L499)) Undefined variable 'value'
- [x]  `foqus_lib.framework.optimizer.SM_Optimizer:L507` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/optimizer/SM_Optimizer.py#L507)) Undefined variable 'value'
[...]

PR #806, opened 2020-09-15
pylint: add explicit imports to get rid of undefined symbol errors

lbianchi-lbl:
Fixes pylint errors described in #805 . I think that the addition of the explicit imports from `pyomo.environ` shouldn't have any effect on the runtime behavior, but it's of course possible that there's some known issue or any other gotchas that I'm not aware of.


anujad95 (commented):
The following imports have been tested with one of the SM Optimizer examples in the repo, and it seems to be working:
    from pyomo.environ import (Var,
                               Constraint,
                               ConstraintList,
                               ConcreteModel,
                               PositiveReals,
                               value,
                               Objective)
Please modify the imports accordingly. Thanks.

lbianchi-lbl:
@anujad95 thanks for looking into this. I've added the missing imports as per your comment.

Just one quick question: if I understand correctly, the extra imports in 52079df were needed because the code didn't run with only the imports in afe2766? The reason I'm asking is because pylint didn't detect other import errors after afe2766. If these imports do cause errors, this would be a false negative (real error which is ignored by pylint), and we could think of adding a test to cover this.

anujad95:
@lbianchi-lbl, Warmstart is not an object within the pyomo package, rather, it is a keyword argument. So if that is included in the imports, since it does not exist, and it is mentioned within the try, except block, when FOQUS opens, it will show that SM optimizer plugin is unavailable due to missing dependency. So, Warmstart needs to be removed. 

The extra objects Constraint, ConcreteModel and PositiveReals might not be required since they are already included in a the pyomo model plugin that gets imported in the SM_Optimizer file. However, i have added it just to be on the safer side.

lbianchi-lbl:
Oh, I see now, thanks! So, if I understand correctly, `Warmstart` has nothing to do with Pyomo? I assumed that all the missing variables that pylint was complaining about came from Pyomo imports, but in retrospective there's no real reason why it should be so.

I've now removed the wrong import, but at this point the `Warmstart` being assigned to `kwds['warmstart']` [here](https://github.com/CCSI-Toolset/FOQUS/blob/59682487d92e88a59fb620a1d62e67339113c548/foqus_lib/framework/optimizer/SM_Optimizer.py#L428) could be a legitimate missing variable, since I couldn't find it anywhere else in the module. On the other hand, I see that there's a lot of `eval()`ing going on, so maybe `Warmstart` is defined in whatever snippet is being loaded at runtime there.

@anujad95 do you know have more info about where that `Warmstart` could be coming from?

anujad95 (approved):
LGTM

Issue #807, opened 2020-09-15
pylint errors in foqus_lib/framework/listen/listen.py


lbianchi-lbl:
Parts of the efforts towards #600.

- [ ]  `foqus_lib.framework.listen.listen:L58` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/listen/listen.py#L58)) Undefined variable 'logging'
- [ ]  `foqus_lib.framework.listen.listen:L65` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/listen/listen.py#L65)) Undefined variable 'logging'

These seems to be due to a simple missing import from the standard library.

PR #808, opened 2020-09-15
pylint: add missing import

lbianchi-lbl:
This should be a straightforward fix for the missing import caught by pylint described in #807.


ksbeattie (approved):
LGTM

PR #814, opened 2020-09-15
Refactor method args to get rid of pylint false-positive errors

lbianchi-lbl:
Part of the efforts towards #600.

- [x]  `foqus_lib.framework.uq.Model:L275` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/uq/Model.py#L275)) Method has no argument
- [x]  `foqus_lib.framework.uq.Model:L296` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/uq/Model.py#L296)) Method has no argument

These errors are false positives, in the sense that the code works, but the slightly unusual syntax seems to annoy pylint. The changes in this PR are a possible solution for using a more idiomatic syntax that also manages to make pylint happy.

### DIFF
--- a/foqus_lib/framework/uq/Model.py
+++ b/foqus_lib/framework/uq/Model.py
@@ -272,9 +272,7 @@ class Model:
     def getNamesIncludeNodes(self):
         return self.namesIncludeNodes
 
-    def setInputNames(*arg):
-        self = arg[0] #Reference to self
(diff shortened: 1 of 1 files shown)


ksbeattie (approved):
Hopefully this will be ok!

Issue #815, opened 2020-09-15
pylint errors in foqus_lib/framework/uq/SampleRefiner.py


lbianchi-lbl:
Part of the efforts towards #600.

- [ ]  `foqus_lib.framework.uq.SampleRefiner:L25` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/uq/SampleRefiner.py#L25)) Undefined variable 'QMessageBox'
- [ ]  `foqus_lib.framework.uq.SampleRefiner:L26` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/uq/SampleRefiner.py#L26)) Undefined variable 'QMessageBox'
- [ ]  `foqus_lib.framework.uq.SampleRefiner:L26` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/uq/SampleRefiner.py#L26)) Undefined variable 'QMessageBox'

This seems to be a straightforward missing import for `QMessageBox`.

PR #816, opened 2020-09-15
Add missing PyQt5 import in foqus_lib/framework/uq/SampleRefiner.py

lbianchi-lbl:
This should address the missing import found by pylint as described in #815.


ksbeattie (approved):
LGTM

PR #820, opened 2020-09-16
Address PyQt-related errors when initializing GUI in foqus_lib.foqus

lbianchi-lbl:
This PR addresses PyQt5 errors occurring when invoking `foqus` in GUI mode as described e.g. in #821. Although it was not possible to reproduce the error consistently, the cause seems to be related to the way PyQt5 modules are imported in `foqus_lib.foqus`. This PR adds an explicit import of the required PyQt5 modules in the `guiImport()` function, to make sure that they're properly initialized when the GUI objects are instantiated.

In addition to the changes to the code, I added two tests. Although the conditions under which the original error occurs aren't known precisely, the tests will fail in similar circumstances.

1. Check that it's possible to create a `QApplication` instance after running the `foqus_lib.foqus.guiImport()` function
1. Test that the actual FOQUS GUI window is created when running `foqus_lib.foqus.main` with the appropriate CLI arguments
  - To complete the test, the GUI window must be closed manually
  - This is considered temporary: in the long run, a better automated solution should be used. In the meantime, to avoid causing issues in automated test environments (such as CircleCI) the test is marked so that it will only run if pytest is invoked with the `-m requires_human` option
  - Alternatively, if we decide that the first test is enough, we could get rid of the manual test altogether

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,10 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install --progress-bar=off -r requirements-dev.txt
+      - run:
+          name: Install system packages needed to run GUI-related PyQt tests
+          command: |
+            apt update && apt install -y libxkbcommon-x11-0 x11-utils
 
       - persist_to_workspace:
           root: ~/repo
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,10 +16,6 @@ jobs:
             python3 -m venv venv
             . venv/bin/activate
             pip install --progress-bar=off -r requirements-dev.txt
-      - run:
-          name: Install system packages needed to run GUI-related PyQt tests
-          command: |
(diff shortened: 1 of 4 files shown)


lbianchi-lbl:
CircleCI builds are failing.

- I believe the reason is that running tests using Qt GUI functionality requires extra system dependencies (on Debian/Ubuntu `libxkbcommon-x11-0` and possibly `x11-utils`)
-  I've added an extra `apt-get` step to the `setup` job in .circleci/config.yml`, which works fine, but the changes are not available in the subsequent `test` job because (I think) it starts from the original Docker image
- If this is true, to be able to run these tests, we'd need to add the system dependencies to the `test` environment, e.g. by adding the `apt-get` step to the Docker image in some way

ksbeattie:
Waiting on a test from @mirandamartin on if this does indeed fix the problem she is seeing.

ksbeattie (approved):
LGTM

Issue #772, opened 2020-07-22
Addressing usability study findings


sotorrio1:
After Sarah completes her usability study for SDoE, we will need to address any changes she recommends/suggests according to her findings.

ksbeattie:
Sarah says she should have some design suggestions for us in mid-Aug

ksbeattie:
@sotorrio1 says he's in the process of getting changes from the LANL team to address the simpler/light-weight/easier/quick solutions for this Sept release.

ksbeattie:
PR #823 above addresses the "quick" solutions.  This issue will stay to track any of the more involved GUI improvements.

PR #823, opened 2020-09-23
SDoE: Addressing usability study findings

sotorrio1:
Addressing usability study findings. Fixes issue #772


mirandamartin (commented):
Things look really good, Pedro. Just a couple of things. 
My biggest comment is that the tooltip for MWR in the Desired Design Size and NUSF Weights box didn’t pop up for me at all. 
The other thing I'll mention isn't a must-happen-now fix but a if-possible-soon fix. The bulleting and numbering in the paragraph in the Design Setup box is a little off. We meant to have an indent for the lines you numbered 1 and 2 so that it looks like it's within that bullet point. If an indent isn't possible, change lines numbered 1 and 2 to have dashes in the beginning, and removing all the bullets from the other lines would work as well. 
I'm going to also send an email with all of this info (and a visual of what I just described) so hopefully it all makes sense.

sotorrio1:
@mirandamartin Have you hover over the label 'MWR' before the drop-downs for the MWR values? From the slides it looked you wanted the tooltip in the MWR label. If you want the tooltip for the whole group box I can change that. 
Regarding the indent, let me see what I can do.

mirandamartin:
@sotorrio1 Oh I see what you mean. Yes I just tried it when hovering over the label 'MWR' and it worked. I think it might be better to see it when hovering over the box (or both the boxes and the label?). The powerpoint isn't clear on that, sorry!

candcook (approved):

candcook:
This is looking good - I think the useability is greatly improved.

PR #824, opened 2020-09-23
Some fixes for UQ/OUU demo

sotorrio1:
Making sure UQ and OUU work smoothly for the upcoming demo


sotorrio1:
@anujad95 Just make sure to test the following: (Loading the Rosenbrock session would be enough)

1. **UQ:**

- [ ] Make sure you can toggle between "Wizard" and "Expert" mode in the UQ Analysis window.

- [ ] Make sure the 3-input vs 1-output visualization for Response Surface works.

2. **OUU:**

- [ ] Make sure OUU runs 

- [ ] Make sure after OUU finishes, you can scroll down the plots area and FOQUS doesn't crash.

sotorrio1:
@jmorgan29 Would you mind taking a look at this PR? You only need to test what I mentioned to Anuja in the previous comment. Thanks!

anujad95 (approved):
I tested the rosenbrock example for everything that Pedro mentioned, and it works fine.

Issue #829, opened 2020-10-07
Issue when plotting a new design with historical runs included


Daison2102:
Issues when reproducing Examples USF-2 in the Sequential Design of Experiments.

This issue is related to plotting a new design with the historical runs included. For the example USF-2, I used the attached files, the “SDOE_Ex1_Candidates.csv” file as candidate file and the “usf_d8_n10000_Label+X1+X2.csv” as history file. The options for the SDOE were: Uniform Space Filling, Minimax, Min Design Size = Max Design Size = 8, Label Column = index and X1, X2 = inputs.
After the SDOE module created the design, I tried to plot the new design along with the historical data by selecting the View option in the Created Designs Table, and then selecting Plot SDOE. The issue below arises in the terminal at this point.  The same issue arises when reproducing example Example USF-3.

Traceback (most recent call last):
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoePreview.py", line 118, in plotSdoe
    plot_utils.plot(fname, scatterLabel, hname=hname, show=show, nusf=nusf)
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\framework\sdoe\plot_utils.py", line 149, in plot
    df, hf = load_data(fname, hname)
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\framework\sdoe\plot_utils.py", line 40, in load_data
    assert (names == list(hf))
AssertionError

- **Files**

[files.zip](https://github.com/CCSI-Toolset/FOQUS/files/5343632/files.zip)

![Issue when plotting a new design with historical runs included](https://user-images.githubusercontent.com/32642037/95384252-5355c000-08ba-11eb-86ab-7828db9c1208.png)

sotorrio1:
This issue and issue #828   should get fixed with PR #831

Daison2102:
Fixed

Issue #828, opened 2020-10-07
Issue using candidate and history file in SDOE


Daison2102:
Issues when reproducing Examples USF-1 and USF-2 in the Sequential Design of Experiments.

The generated files for the selected designs for the example USF-1, don’t have the Label column when using the version of FOQUS 3.6.0 (see “usf_d8_n10000_X1+X2.csv” in the attached zip file), while that using FOQUS 3.5.1, the generated files for the selected designs do have the Label column (see “usf_d8_n10000_Label+X1+X2.csv” in the attached zip file). This generates an issue when reproducing example USF-2, which use the “SDOE_Ex1_Candidates.csv” as candidate file and the design from example USF-1 as history file (see the attached picture). The issue arises when clicking on the Confirm button to advance to the Ensemble Aggregation Window. Se details of the issue below. This issue seems to be related to the missing label in the history file, which wasn’t generated by FOQUS 3.6.0. If I use the design file generated by FOQUS 3.5.1 instead, no issue arises.

Error
Traceback (most recent call last):
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoeSetupFrame.py", line 157, in confirmEnsembles
    self.updateAggTable()
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoeSetupFrame.py", line 357, in updateAggTable
    self.updateAggTableRow(0)
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoeSetupFrame.py", line 518, in updateAggTableRow
    candidateData, historyData = self.createAggData()
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoeSetupFrame.py", line 207, in createAggData
    cand_agg, hist_agg = self.aggregateEnsembleList()  # these are dfs
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\gui\sdoe\sdoeSetupFrame.py", line 203, in aggregateEnsembleList
    cand_agg, hist_agg = df_utils.check(cand_csv_list, hist_csv_list)
  File "c:\anaconda3\envs\foqus\lib\site-packages\foqus_lib\framework\sdoe\df_utils.py", line 44, in check
    hist_df = hist_df[cols]
[...]

PR #831, opened 2020-10-08
SDoE: Fixing USF sequential design

sotorrio1:
Fixing the issue with USF sequential design reported on issues #828 and #829


Daison2102 (approved):
Fixed

PR #834, opened 2020-10-20
Replace foqusPSUADEClient BAT file with entry point

### LINKED ISSUE #833: OUU: foqusPSUADEClient.bat not included in PyPi package
When trying to run OUU on Windows, FOQUS tries to find `foqusPSUADEClient.bat` but it wasn't included in the PyPi package. 

@ksbeattie suggests not being OS specific and turn `foqusPSUADEClient` into a command that gets installed by PyPi based on the OS at install time, just like the `foqus` command itself. So Mac, Linux and Windows just call `foqusPSUADEClient` and on Mac and Linux that will be a python script, but on Windows it's an executable (.exe)

lbianchi-lbl:
Fixes #833 

## Summary of the changes

- Use entry point script (automatically created when installing the PyPI package) for built-in cross-platform support
  - This obviates the need for the ad-hoc BAT file for Windows compatibility
- Use the full path of the entry point executable for compatibility with PSUADE
  - When loading the various driver options, PSUADE explicitly check that their values correspond to existing files ([link](https://github.com/LLNL/psuade/blob/4576fa16e491509068439b1d15ab3d0cd8c39e65/Src/DataIO/FunctionInterface.cpp#L875-L882))
  - Once installed, the `foqusPSUADEClient` can be run as a command, but is interpreted by PSUADE as a non-existing relative path in the working directory
  - Passing the full path to the `foqusPSUADEClient` executable when invoking PSUADE satisfies the PSUADE check without requiring additional platform-specific procedures

---

(this was originally added as a comment to the WIP PR)

@sotorrio1 this first pass should be enough to create the `foqusPSUADEClient` executable, however there seems to be some functionality that still depends on the `foqusPSUADEClient.bat` file being present that I'm not sure how to treat, e.g. in `foqus_lib.gui.ouu.ouuSetupFrame.ouuSetupFrame.setupPSUADEClient()` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/123fc15687d28cde9d0ce05bd1b9392a4580f919/foqus_lib/gui/ouu/ouuSetupFrame.py#L739-L740)).


sotorrio1:
@lbianchi-lbl if your entry point works, it would be just a simply command `foqusPSUADEClient` correct? We might just be able to delete the `setupPSUADEClient()` method and point to the entry point in `foqus_lib.gui.ouu.ouuSetupFrame.ouuSetupFrame.analyze()` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/123fc15687d28cde9d0ce05bd1b9392a4580f919/foqus_lib/gui/ouu/ouuSetupFrame.py#L908)).

lbianchi-lbl (commented):
@sotorrio1 something along these lines?

sotorrio1:
@lbianchi-lbl yes, I guess something like that. Let me test and I'll get back to you :)

lbianchi-lbl:
Thanks to @sotorrio1, I was able to test this locally and reproduce the issue that he had with the strategy currently implemented with this PR.

Even though the `foqusPSUADEClient` executable is correctly installed in the environment (as per `which foqusPSUADEClient`), it's not found by PSUADE:

```txt
**********************************************************************
*      Welcome to PSUADE (version 1.7.10)
**********************************************************************
PSUADE - A Problem Solving environment for
         Uncertainty Analysis and Design Exploration (1.7.10)
(for help, enter <help>)
======================================================================
psuade> readApplication WARNING: opt driver foqusPSUADEClient not found.
readApplication WARNING: ensemble opt driver foqusPSUADEClient not found.
**********************************************************************
```

For reference, the steps I used for testing are:

- Open FOQUS
- Open a session…
- Go to ‘examples/test_files/UQ/Rosenbrock.foqus’ and load it
- Once the Flowsheet model is loaded go to OUU tab
- Select “Choose Flowsheet Node” option and in the dropdown select “Rosenbrock” (there is only one node in this model).
- Then you need to pick the variable types. For this example just do: x1, x2 and x3 as Z1 (primary opt variables) and x4, x5, and x6 as Z4 (continuous UQ variables).
- Leave the defaults for the first tab in the bottom, make sure Rosenbrock.f is your objective, and in UQ setup tab choose generate sample for Z4 (let’s do 10 samples Latin Hypercube).
- In the last tab run OUU. 

---

I'll now try to go back to the previous strategy and test how it goes when generating the BAT file programmatically on Windows.

lbianchi-lbl:
@sotorrio1 this method lets the entry point be compatible with PSUADE, and should work regardless of the platform.

At the moment I've tested this on Linux, and should be able to test it on macOS and Windows later this week; if you get a chance to check it before that, feel free to update the checklist:

- [ ] Linux (Ubuntu 20.04.1)
- [ ] macOS
- [ ] Windows 10

sotorrio1:
@lbianchi-lbl just tested on Mac and it works (just updated the list). I'll try to test on Windows ASAP and let you know.

lbianchi-lbl:
@sotorrio1 great, thanks for the tests!

I believe this new error is unrelated, and most likely caused by the trailing `\r` in the string (on Windows, line endings are `\r\n` as opposed to `\n` in Unix systems). A quick fix to try would be to use `str.strip()` inside `float()`:

```py
x.append(float(regex[0].strip()))
```

Let me know if this works (if you could. maybe in a new issue/PR?)

sotorrio1 (approved):
The entry point solution works. Opened a new issue for the regex problem

sotorrio1:
@lbianchi-lbl I'm sorry, I think I messed up. While testing on Windows the FOQUS working directory was pointing to a previous version of FOQUS. When I realized that I updated the wd to the correct one and the entry point on Windows is not working. I get the error when you do the assert "The 'foqusPSUADEClient' executable was not found"

lbianchi-lbl:
@sotorrio1 no worries! I think the first thing to do would be to see if it's a problem of the environment in general, or of the `foqusPSUADEClient` setup in this PR. Can you try the following:

- From the shell (cmd.exe, Anaconda shell, Powershell, ...) where `foqus` is started (or, a shell in the same Conda env):
  - `where.exe foqus`
  - `where.exe foqusPSUADEClient`
  - `where.exe` is similar to `which` on Unix. In case it doesn't work, some alternatives to try are described here: https://superuser.com/questions/207707/what-is-windows-equivalent-of-the-which-command-in-unix-is-there-an-equivale
- From a Python shell in the same Conda environment:

  ```py
  >>> import shutil
  >>> shutil.which('foqus') or 'foqus not found'
  >>> shutil.which('foqusPSUADEClient') or 'foqusPSUADEClient not found'
  ```

sotorrio1:
@lbianchi-lbl in both cases (terminal and python) foqusPSUADEClient can't be found, but foqus is found.

ksbeattie (dismissed):
me likey

ksbeattie:
@sotorrio1 will test this on Windows in a clean conda env

sotorrio1:
@lbianchi-lbl finally tested in a fresh environment and the entry point works on Windows as well. I ran into an issue though, see below:
Screen Shot 2020-11-17 at 12 07 45 PM

lbianchi-lbl:
OK, it seems I managed to fix the Windows issues that were haunting me and I was able to successfully test that the setup implemented in this PR works. The steps I used where the ones provided by Pedro, with the addition of setting the path to the PSUADE executable in the Settings tab.

Without doing that, a popup window would appear after hitting "Run", but, instead of showing the "PSUADE not found" error message, it was blank and unresponsive, eventually causing the whole window to freeze and requiring the application to be terminated. Since I've experienced other similar issues on the same Windows installation on other Python projects as well, I think this "error within the error" crashing behavior might be specific to my system and not related to FOQUS. 

To verify that the correct executable was picked up by PSUADE, I checked that it is present in the content of the `OUU_files/ouuTemp.ouudat` file, under the `opt_driver` and `ensemble_opt_driver` keys:

```txt
PSUADE_IO
6 1 1
1 0
  4.0000000000000000e+00
  5.0000000000000000e+00
  4.0000000000000000e+00
  5.0000000000000000e+00
  4.0000000000000000e+00
  4.0000000000000000e+00
 9.9999999999999997e+34
PSUADE_IO
PSUADE
INPUT
   dimension = 6
   variable 1 Rosenbrock.x1  =  -1.0000000000000000e+01   1.0000000000000000e+01
   variable 2 Rosenbrock.x2  =  -1.0000000000000000e+01   1.0000000000000000e+01
   variable 3 Rosenbrock.x3  =  -1.0000000000000000e+01   1.0000000000000000e+01
   variable 4 Rosenbrock.x4  =  -1.0000000000000000e+01   1.0000000000000000e+01
   variable 5 Rosenbrock.x5  =  -1.0000000000000000e+01   1.0000000000000000e+01
   variable 6 Rosenbrock.x6  =  -1.0000000000000000e+01   1.0000000000000000e+01
END
OUTPUT
   dimension = 1
   variable 1 Rosenbrock.f
END
METHOD
   sampling = MC
   num_samples = 1
random_seed = 41491431
END
APPLICATION
   driver = NONE
   opt_driver = C:\Users\lbian\MINICO~1\envs\CCSI-F~1\Scripts\FOQUSP~1.EXE
   ensemble_opt_driver = C:\Users\lbian\MINICO~1\envs\CCSI-F~1\Scripts\FOQUSP~1.EXE
   aux_opt_driver = NONE
   launch_interval = 0
END
ANALYSIS
   optimization method = ouu
   optimization num_local_minima = 1
   optimization max_feval = 1000000
   optimization fmin = 0.0
   optimization tolerance = 1.000000e-06
   optimization num_fmin = 1
   optimization print_level = 3
   analyzer output_id = 1
   opt_expert
   printlevel 0
END
END
```

PR #835, opened 2020-10-20
Make import guard for dmf_lib more specific and disable pylint import-error check

lbianchi-lbl:
This PR is part of the ongoing effort for addressing pylint false-positive errors towards #600.

In this case, the imports within the import guard for `dmf_lib` (`try import ... ; except:`) are reported as `import-error`s by pylint, even though the possible missing import is dealt with correctly at runtime.

The changes in this PR are almost completely pylint-specific: the only modification to the runtime behavior is to use a more specific exception type in the `except` clause, as this is usually the preferred way to implement this type of functionality.

### DIFF
--- a/foqus_lib/gui/model/dmfUploadDialog.py
+++ b/foqus_lib/gui/model/dmfUploadDialog.py
@@ -11,6 +11,7 @@ import logging
 import platform
 import subprocess
 try:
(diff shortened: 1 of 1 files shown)


ksbeattie (approved):
LGTM

PR #836, opened 2020-10-20
Add more imports for surrogate modeling

anujad95:
The log and exp functions are used from pyomo imports. With the change in pyomo environ import style, these additional imports are required.


ksbeattie (approved):
LGTM

lbianchi-lbl (approved):
Looks good!

Adding a note for the future that we should find a way to cover this in a test, since potential errors of this type are impossible for pylint (or other static analysis tools) to detect.

Issue #840, opened 2020-10-27
py2 code in foqus_worker


boverhof:
# Description
The last commit to this code was in 2018 and was titled '2to3', so the converted didn't catch this exception synatx.

```
(py3) joshuas-mbp:foqus-boverhof boverhof$ foqus_worker.py 
Traceback (most recent call last):
  File "/Users/boverhof/Documents/workspace.ccsi2/py3/bin/foqus_worker.py", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/cloud/aws/foqus_worker.py", line 279
    raise NotImplementedError, 'Flowsheet Node model type: "%s"' %(str(dat.flowsheet.nodes[nkey].turbApp))
                             ^
```
# Solution
Remove this script, I don't believe it is used anymore.

PR #841, opened 2020-10-27
https://github.com/CCSI-Toolset/FOQUS/issues/840

boverhof:


ksbeattie (approved):
LGTM, though I'd also like to get @lbianchi-lbl to look at this as well, as we were talking about possible changes to setup.py in another context.

lbianchi-lbl (approved):
Looks good. I `grep`ped for `foqus_worker` to be extra sure that there aren't any dangling references left, but the `scripts` entry in `setup.py` is the only one.

PR #832, opened 2020-10-13
Ensure that all R files are installed

lbianchi-lbl:
Should fix #827.

## Changes

In `setup.py`:

- Add `*.R` pattern to `package_data`
- Use `find_namespace_packages()` instead of `find_packages()` so that also files inside directories without `__init__.py` are included as well

PR #849, opened 2020-11-17
Revert to find_packages() to find packages

lbianchi-lbl:
## Motivation

- In #832 we switched from `find_packages()` to `find_namespace_packages()`in `setup.py`. This allows non-Python files to be installed without having to create a `__init__.py` file in the directory where they are located
- However, this had the unintended consequence of directories other than `foqus_lib`, such as `examples/` and `cloud/`, to be included in the in the Python distribution package, which in turn resulted in multiple errors showing up in e.g. CI checks
- Moreover, even though it would be possible to address these particular issues, our use case falls outside of the primary use case for `find_namespace_packages()`, which is to create multi-distribution packages (i.e. Python packages that are distributed and installed separately, but operate as a single package at runtime)

## Solution

- Revert to using `find_packages()` to find packages in `setup.py`, and add `__init__.py` files wherever necessary
- This is the simplest (and therefore recommended) solution for the time being. @ksbeattie and I are also looking into more long-term solutions to streamline the `setup.py`/packaging configuration

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 John Eslick, Carnegie Mellon University, 2014
 See LICENSE.md for license and copyright details.
 """
-from setuptools import setup, find_namespace_packages
+from setuptools import setup, find_packages
 import sys
 import os
 import subprocess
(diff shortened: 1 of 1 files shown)


lbianchi-lbl:
Looks like the previously failing Appveyor check, which tests `python setup.py install` on Windows, is now successful.

ksbeattie (approved):
LGTM

PR #861, opened 2020-12-17
Shorten directory names in `examples` to avoid filename length limitations on Windows

lbianchi-lbl:
If any of the file names exceeds the maximum allowed filename length on Windows, multiple operations will fail on Windows, including cloning the repository or installing with pip from a GitHub URL.


ksbeattie (approved):
LGTM

PR #863, opened 2020-12-20
SM Optimizer Improvements

anujad95:
The following changes were made to the SM optimizer:
1. Simplified multi start initialization at each iteration
2. Time taken to implement each step according to the algorithm, in each iteration is displayed on the message window
3. Restructured solver settings GUI - math optimization solver options can be specified by the user in a dictionary, within the GUI
4. Minor code style changes to follow the pep 8 standards as much as possible


Daison2102 (approved):
Tested with several pyomo and gams solvers (ipopt, conopt and knitro) with different options for each solver and SM optimizer worked fine. Also, I double checked the multi start initilization part and it looks good to me

PR #868, opened 2021-01-21
Implementation of robust Optimal Design of Experiments (ODoE)

sotorrio1:
Adding the latest feature to the SDoE module. User will need psuade-lite to use it.


candcook (dismissed):
This is looking good.

ksbeattie:
@sotorrio1, should we wait until psuade-lite is ready and open to merge this, or can we safely do it before without requiring psuade-lite that isn't yet available?

candcook (approved):
Looking good

PR #862, opened 2020-12-17
pylint fixes (5 of 7) from #812

ksbeattie:
I (think) I've fixed 5 of the 7 pylint issues from #812.  They really looked like typos.

@boverhof, please review these and suggest (or feel free to commit to this branch) fixes for the last 2.

Issue #812, opened 2020-09-15
pylint errors in foqus_lib/framework/sim/turbineConfiguration.py


lbianchi-lbl:
Parts of the efforts towards #600.

- [x]  `foqus_lib.framework.sim.turbineConfiguration:L995` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L995)) No value for argument 'jobID' in method call
- [x]  `foqus_lib.framework.sim.turbineConfiguration:L1004` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1004)) Too many arguments for format string
- [x]  `foqus_lib.framework.sim.turbineConfiguration:L1007` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1007)) Undefined variable 'jobid'
- [x]  `foqus_lib.framework.sim.turbineConfiguration:L1010` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1010)) Undefined variable 'jobid'
- [ ]  `foqus_lib.framework.sim.turbineConfiguration:L1229` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1229)) Undefined variable 'sinterConfigPath'
- [ ]  `foqus_lib.framework.sim.turbineConfiguration:L1236` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1236)) Undefined variable 'sinterConfigPath'
- [x]  `foqus_lib.framework.sim.turbineConfiguration:L1454` ([link](https://github.com/CCSI-Toolset/FOQUS/blob/master/foqus_lib/framework/sim/turbineConfiguration.py#L1454)) errList is not callable

lbianchi-lbl:
Just as a heads-up, the last error in the list ("errList is not callable") might be a pylint false positive. I'll look into that.

ksbeattie:
@boverhof these all look pretty simple to fix.  Can you take a look?

PR #871, opened 2021-02-02
pylint fixes home stretch

lbianchi-lbl:
- The changes in this PR address all remaining pylint errors, and switches on the actual pylint check in CircleCI
  - This does not include the errors in #812 being addressed in #862 (i.e. we need to merge in #862 to have zero pylint errors on the whole codebase)
- The fixes address both false positives (functioning code mistakenly flagged as error by pylint) and true positives (pylint errors that would cause actual errors at runtime)
  - Some fixes for true positives silence the pylint error but do not address the error itself; these are marked with `# TODO pylint: disable=...` (as opposed to the standard `# pylint: disable=...`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,7 @@ jobs:
           name: PyLint
           command: |
             . venv/bin/activate
-            pylint --rcfile=.pylint/pylintrc --disable=W,C,R,I foqus_lib || true
+            pylint --rcfile=.pylint/pylintrc --disable=W,C,R,I foqus_lib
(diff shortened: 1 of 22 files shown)


sotorrio1 (commented):
LGTM

boverhof (approved):
Looked over the changes and think this can be accepted.

ksbeattie (approved):
This has been a long time coming.  A little nervous about all the "this code must not be getting executed" parts but having pylint enabled now, I hope will be worth it.

PR #876, opened 2021-02-09
Vector Variable Support - Python

anujad95:
This has the required code changes for supporting python model based vector variables. It would still need some refinement and modifications.  This addresses #580.

PR #883, opened 2021-02-25
Add vectors (DO NOT MERGE)

eslickj:
Just a simple half done concept for vectors.  Just made the PR to share the idea.  

You can add input vectors by entering the name as something like ```x[1,2]```  strings for indexes should be okay.  Output vector would work the same way but I didn't bother finishing the GUI update for output.  A attached a session file demo.  This isn't really ready, just wanted to share the simplest possible vector implementation that would seamlessly integrate with what we have.  It's sort of a weird, but I wanted to piggy-back on the existing variable list without changing any existing interface. 

[test.zip](https://github.com/CCSI-Toolset/FOQUS/files/6046223/test.zip)

### DIFF
--- a/foqus_lib/framework/graph/nodeVars.py
+++ b/foqus_lib/framework/graph/nodeVars.py
@@ -45,6 +45,39 @@ class NodeVarListEx(foqusException):
         self.codeString[7] = "Var name already in use, cannont add"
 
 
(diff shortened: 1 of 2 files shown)


ksbeattie:
This is an experimental work in conjunction with PR #876

PR #896, opened 2021-03-16
Disable pylint check for import error when running module as script

lbianchi-lbl:
- This addresses is a minor issue that causes the pylint checks to fail
- As this is a pylint-only fix, it doesn't affect the runtime behavior, and therefore is safe to merge in
- The only question I had was to ask more details about how the script mode (`if __name__ == '__main__')` and the `import user_plugins` line work in practice


ksbeattie (approved):
I'm not sure why there is that check for '__main__'.  I'm not aware that this would be run as a stand-alone script.  I'm thinking we make and issue to ask @anujad95  about that later, and merge this PR in the meantime.

Issue #897, opened 2021-03-16
`import user_plugins` in SM_Optimizer.py?


ksbeattie:
In PR #896 @lbianchi-lbl and I were curious if [here in SM_Optimizer.py](https://github.com/CCSI-Toolset/FOQUS/blob/0dd0f74225c1034eeb8bc5118b98bf5c8be0b051/foqus_lib/framework/optimizer/SM_Optimizer.py#L36) the `if __name__ == '__main__'` and the `import user_plugins` is needed still?

anujad95:
@ksbeattie, actually this import statement can be removed, since line 343 in the SM_Optimizer.py code takes care of it. Thanks for pointing it out.

PR #898, opened 2021-03-24
Removing import statement from SM_Optimizer.py

anujad95:
This should take care of #897 
I tested a tutorial example and it works fine.


lbianchi-lbl (approved):
Thanks @anujad95, this looks good.

Issue #889, opened 2021-03-09
update images in read the docs manual


ksbeattie:
Quite a few of the screenshot images in the RTDs are either out of date with recent releases or of low quality.

For example:
- the first few images in the [ALAMO tutorial](https://foqus.readthedocs.io/en/stable/chapt_surrogates/tutorial/alamo.html)

These should be updated.

PR #900, opened 2021-04-20
Get RTDs to use full width of page

ksbeattie:
This should stop (most of) the images in the docs from being resized (on large monitors) as described in PR #889 

Regardless, I think it looks better to use the full width of the browser screen for displaying docs, though I'd like _others to review how this looks_ as it changes **EVERY PAGE** of our docs!

I also just added to all PRs (via a change to the RTDs config) that the docs are built for that PR and a link to that is below as part of all the other tests that are run.

### DIFF
--- a/docs/source/_static/schema.css
+++ b/docs/source/_static/schema.css
@@ -96,3 +96,7 @@ html {
     color: black;
     font-weight: 500;
 }
+
+.wy-nav-content {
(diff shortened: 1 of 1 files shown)


sotorrio1 (approved):
It looks way better this way. Me likey 👍

eslickj (approved):

Issue #882, opened 2021-02-25
Update year in copyright


ksbeattie:
Should be 2021 and not 2020 or 2019.

Both in the top-level docs and in each file.

PR #906, opened 2021-05-12
Adding copyright/license headers

ksbeattie:
This is a huge PR, it touches nearly every file, addressing #882 

Primarily this adds a common copyright and license notification to the top of every .py file, and a few other kinds of files.

Secondarily, it also:
  - Changes the few files that had DOS-style (`\r\n`) line ending to have unix-style `\n` line endings, so things are consistent
  - Removed the execute permissions on files changed
  - Added a newline to the end of files that were missing them


foqus_lib/framework/optimizer/BFGS.py
```diff
@@ -1,3 +1,18 @@
...
+# Austin, URS Energy & Construction, Inc., et al.  All rights reserved.
+#
+# Please see the file LICENSE.md for full copyright and license information,
+# respectively. This file is also available online at the URL
+# "https://github.com/CCSI-Toolset/FOQUS".
+#
+###############################################################################
 """#FOQUS_OPT_PLUGIN BFGS.py
 
 Optimization plugins need to have #FOQUS_OPT_PLUGIN in the first 150 characters,
```
lbianchi-lbl, line 18: The failing tests are because of a `KeyError` with `BFGS` not being in the `optPlugins` dictionary of the `session`/`self.dat` instance. Maybe this requirement to identify plugins was inadvertently broken by the added header, causing the BFGS plugin to be ignored, and in turn the tests to fail?

ksbeattie:
@eslickj, @anujad95, might there be a better way to detect plugins than parsing the code for a specific string?  I'm thinking trying an import and seeing if it succeeds and implements a known interface (or type).

sotorrio1 (commented):
Regarding files within UQ, OUU and SDoE modules; everything looks good to me.

foqus_lib/framework/optimizer/BFGS.py
anujad95: @lbianchi-lbl I am looking into this, and will follow up soon.

eslickj:
@ksbeattie, I think you are right there is likely a better way.  They reason I checked for that comment string was that I was afraid of importing some random py file that did something odd.  If we have a plugin structure where we can be sure only plugin files should be (or at least python files that import without side-effects), we could probably do something better.  I'm afraid we may currently allow users to just drop plugin files anywhere in their working directory, but I can't remember.

anujad95 (approved):
LGTM

lbianchi-lbl (approved):
LGTM :+1:

anujad95 (approved):

PR #909, opened 2021-05-18
Running black on everything

ksbeattie:
Just to see if any of our GHA tests fai.


ksbeattie:
For reference:
- pylint score before black: `4.78/10`
- pylint score after black: `5.19/10`

lbianchi-lbl:
The same things is being done in #975 so this can be closed.

PR #910, opened 2021-05-25
Make warnings failures for ReadTheDocs builds

ksbeattie:
This is a small change so that ReadTheDocs will fail if there are any warnings found when generating our documentation.

Adding a .readthedocs.yaml file to set a config option for RTDs to fail if there are any warnings when generating the documentation.  Adding a similar change to the Makefile so the same happens when building the docs locally.


lbianchi-lbl (approved):
Looks good! :+1:

Issue #880, opened 2021-02-23
Update known issues in FOQUS docs/debugging section


eslickj:
We need to update the know issues section.  Some issues are out-of-date, and there is a recently discovered issue with calculator blocks that use Excel not working properly through the COM interface, only through the GUI interface.

ksbeattie:
@eslickj, any news on this?

eslickj:
Was on vacation.  I'll do it now before I forget.

eslickj:
@anujad95 or @sotorrio1,  Is this from the know issues section still true?:

> The user regression analysis features, iREVEAL and ALAMO, of the UQ
   tool requires a separate Python 2.7 installation. Furthermore, Python
   must be both in PATH variable and associated with .py files. Details
   on installing Python and fixing any issues encountered may be found
   in the iREVEAL Installation Guide and the iREVEAL User Manual, Known
   Issues section.

anujad95:
@eslickj, I don't think Python 2.7 is required for the user regression analysis feature, at least for ALAMO. We can use higher versions of Python. I'm not sure about iREVEAL.

sotorrio1:
As far as I can tell, there is no requirement within the UQ module for python 2.7. Everything should work with python 3. At the same time, I'm not very familar with iREVEAL or ALAMO...

PR #911, opened 2021-06-02
Update known issues

eslickj:
Addresses issue #880.

We should probably run down the list of know issues at the next meeting.  I'm not really up-to-date anymore, so I don't really know what's current.  I removed one issue related to the old py2exe installer, and added the Aspen Plus issue with Excel calculator blocks.


anujad95 (approved):
LGTM

PR #920, opened 2021-06-23
Use INFO level if failing to import optional pyomo dependency in SM_Optimizer plugin

lbianchi-lbl:
## Motivation

- pyomo is an optional dependency for the `SM_Optimizer` plugin, but unlike other plugins, the message reporting the unavailability of plugin dependencies is logged as an error
- This causes a false-positive error (FOQUS doesn't crash), which could nonetheless cause some tests to fail (e.g. if the tests look at log output to check for `ERROR` messages) in addition to being confusing

## Changes

- Use `logger.info()` instead of `logger.exception()` to display error message

### DIFF
--- a/foqus_lib/framework/optimizer/SM_Optimizer.py
+++ b/foqus_lib/framework/optimizer/SM_Optimizer.py
@@ -78,7 +78,7 @@ try:
     from smt.sampling_methods import LHS
     packages_available = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


lbianchi-lbl:
@anujad95 does this make sense? I checked the other plugins and the `ImportError`s for optional dependencies are all logged as `INFO`, so I figured that we could do the same here to avoid having a single `ERROR`-level message for `SM_Optimizer`

ksbeattie (approved):
Looks ok to me...

PR #922, opened 2021-06-25
Adding missing dunder init files

ksbeattie:
These appear to have been missed in #876


lbianchi-lbl (approved):
Tested that the issue is present on the current main, and is solved when applying the changes from this PR.

PR #926, opened 2021-06-29
Websocket library issue 925

boverhof:
# Description
Websocket client library callback interfaces added an argument wsapp to all callbacks which broke the FOQUS Cloud Notifications.  Simple fix, just added the argument into the callbacks and confirmed it works now.


ksbeattie (approved):
LGTM

PR #916, opened 2021-06-17
Add GUI tests and test framework based on pytest-qt plugin

lbianchi-lbl:
## Summary

- This PR is the first step towards implementing tests targeting the FOQUS GUI that are fully integrated with the pytest framework
- The tests are implementing using the third-party `pytest-qt` framework, plus a set of extensions (tentatively named `pytest_qt_extras`) that is being developed as a higher-level wrapper around `pytest-qt`
- These extensions allow writing GUI tests with a more concise and natural syntax, with the goal of making writing tests for FOQUS more efficient and convenient
- The ultimate goal of this improved test setup is to facilitate extending the FOQUS test suite towards a more complete coverage of the codebase

### A: Verify that the new GUI tests work properly on your machine

- Checkout the code from this PR, navigate to the root of the repository, then:

    ```sh
    # activate Conda env
    conda activate ccsi-foqus  # or create new Conda environment
    # install dev dependencies
    pip install -r requirements-dev.txt
    # install PSUADE
    conda install -c conda-forge -c CCSI-Toolset psuade-lite
    # check that psuade is installed
    psuade --help
    ```
- Run the GUI tests:

    ```sh
    pytest --slowdown-wait=200 -sv -x tests-gui/
    ```

  - This should take a few minutes
    - The `-x` flag makes pytest exit immediately after the first failure
  - The FOQUS main window will be opened and run through the actions defined in the tests "by itself" (i.e. no manual intervention should be necessary)
  - The *action* currently being executed will be displayed as a red rectangle around the action's target widget and a label with the action's name
  - Specify a different value for `--slowdown-wait` (in ms) to run the tests so that they appear slower or faster
- There should now be a `.pytest-artifacts` subdirectory, containing these two subdirectories:
  - `foqus_working_dir`, containing the files produced by FOQUS during the test run
  - `screenshots`, containing the screenshots that were generated automatically before and after each action
  - `snapshots`, containing a set of specialized "data dumps" of a parent widget and all its children, used for debugging issues with locating widgets

PR #936, opened 2021-08-03
Address CI failures due to pip install job timeouts

lbianchi-lbl:
## Motivation

- The `build-docs` and `pylint` jobs fail due to repeated timeouts/network errors at the `pip install` stage
- This seems to be due to a failure in pip's dependency solver, causing pip to download multiple versions of the same package from PyPI in rapid succession, leading to the network errors
- The failing jobs were seen in e.g. #916, but most likely they're unrelated to the PR's content
  - Possibly, but not obviously, related to pip-related errors such as e.g. IDAES/idaes-pse#439

## Changes in this PR

- Adopt the same Conda-based strategy already in use for the `pytest` job to setup the environment for the `build-docs` and `pylint` jobs

### DIFF
--- a/.github/actions/setup-foqus/action.yml
+++ b/.github/actions/setup-foqus/action.yml
@@ -11,6 +11,15 @@ inputs:
 runs:
   using: composite
   steps:
+      # IMPORTANT this requires the Conda env setup to be run before this action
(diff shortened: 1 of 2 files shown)


ksbeattie (approved):
LGTM

Issue #935, opened 2021-07-30
SDOE Created Designs Not Unscaled


mirandamartin:
I created a few designs and these were returned with all variables still scaled to be between 0 and 1 or -1 and 1 depending on the type of design, rather than returned to original units. Subsequently, the plots of created designs had the right axes but the points were squished toward the middle (where they reside in scaled form). 
In my specific case, I used a candidate set of two inputs, one response (or weight if nusf) with ranges given in the screenshot. The designs I got back had ranges that were much smaller. 
<img width="400" alt="Screen Shot 2021-07-30 at 3 36 01 PM" src="https://user-images.githubusercontent.com/55799605/127713902-ffcb4be0-29b3-4b2d-92be-ce49cbaa7d61.png">
<img width="1000" alt="Screen Shot 2021-07-30 at 1 26 56 PM" src="https://user-images.githubusercontent.com/55799605/127716510-62e3d00c-697c-44ea-8cf0-7bf6547361d6.png">
<img width="306" alt="Screen Shot 2021-07-30 at 4 12 56 PM" src="https://user-images.githubusercontent.com/55799605/127716601-1553b7f9-1727-464b-97cb-b3d298004757.png">
<img width="352" alt="Screen Shot 2021-07-30 at 4 13 12 PM" src="https://user-images.githubusercontent.com/55799605/127716621-b529d7e9-bfb4-4164-a96d-66adced115dc.png">

PR #938, opened 2021-08-04
Fixing SDoE scaling

sotorrio1:
The reverse scaling in NUSF and IRSF wasn't working properly. 

- For NUSF, the problem was that the original array was being modified before saving it.
- For IRSF, the problem was that the inversion function for the scaling wasn't being used at all.

Should take care of issue #935


mirandamartin:
@sotorrio1 Sorry I haven't had a chance to look at this yet. I'll work on this first thing tomorrow.

mirandamartin (approved):
Works great! Thanks for all the hard work

PR #941, opened 2021-08-09
Address SyntaxWarnings in Python 3.8

lbianchi-lbl:
- Identity comparisons with literals (`a is 1` as opposed to `a == 1`) cause a `SyntaxWarning` in Python 3.8
- This PR addresses two such cases as reported by @anujad95

### DIFF
--- a/foqus_lib/framework/uq/Plotter.py
+++ b/foqus_lib/framework/uq/Plotter.py
@@ -1076,7 +1076,7 @@ class Plotter:
         except AttributeError:
             return # the selected axis (e.g., colorbar) does not have geometry attribute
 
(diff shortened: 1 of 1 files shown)


sotorrio1 (approved):
LGTM.

This is actually a good catch because (at least for me) when using the previous syntax this capability didn't work (user gets a subplot array of figures and left click zoom into one of the subplots, right click goes back to the subplot array).

On a different note, I noticed a few matplotlib deprecation warnings regarding 'get_geometry' and 'change_geometry'. We might want to take care of those before they turn into errors.

ksbeattie (approved):
LGTM

PR #942, opened 2021-08-09
Remove unneeded coveragepy package

lbianchi-lbl:
- The PyPI package `coveragepy` (as opposed to the "Coverage.py" package, which is named `coverage` on PyPI) is most likely not needed, and may result in some installation incompatibilities as reported by @anujad95


ksbeattie (approved):
LGTM

PR #946, opened 2021-08-18
Fixing typo in LaTeX / PDF project name

ksbeattie:
The project name, used only in the LaTeX / PDF output from ReadTheDocs was "CCIS" and not "CCSI".  This fixes that typo.

This PR also adds to `.readthedocs.yaml` the directive to build all formats of the docs.  It seems we stopped building the PDF versions when that `.readthedocs.yaml` file was first introduced (leaving 3.9.0, as the latest release with PDF docs).  Unfortunately RTD does not produce the PDF format when building for a PR, so we'll have to merge this to see if it works or not.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,3 +8,6 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
     fail_on_warning: true
+
+# Build all formats
+formats: all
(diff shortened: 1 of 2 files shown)


lbianchi-lbl (approved):
LGTM

eslickj (approved):

Issue #948, opened 2021-08-20
SDoE: IRSF not plotting design points


sotorrio1:
There is an issue when generating IRSF designs. The pareto plot shows up but when plotting the individual designs, only the bar plot shows up and the scatter plots appear to be empty. 

The designs are being created and saved in the csv files but the plot is not showing up correctly.

PR #949, opened 2021-08-20
Fixing bug in IRSF

sotorrio1:
There was an issue in IRSF that was applying the reverse scaling function twice, hence the final values were way out of range and the design points weren't showing up in the plot. This should take care of issue #948


mirandamartin:
Looks great!

mirandamartin (approved):
Looks great! Thanks for doing the digging.

Issue #846, opened 2020-11-03
ALAMO plugin for SM validation - error


anujad95:
A plugin named "alamo_surrogate_uq" gets created after the surrogate model is built, meant for uploading in the UQ analyze section for validation against the simulation ensemble. Once the parity plots are generated in the UQ tab, the results are saved, and the FOQUS file is closed, the user is unable to open it again without the alamo_surrogate_uq.py file. This needs to be fixed.

PR #951, opened 2021-08-26
SM UQ Plugin Tutorial

anujad95:
This should take care of #846 
It is something the users must take note of, while opening a FOQUS file containing the response surface validation results in the UQ module. This note has been added to the appropriate tutorial.


ksbeattie (approved):
LGTM

PR #957, opened 2021-09-07
GUI tests documentation and performance improvements

lbianchi-lbl:
- Documentation:
  - Add how to run tests to the developer docs
  - Add stub of dedicated section on how to write tests (including GUI tests)
- Performance: only generate debug dumps with recursive GUI widget information only in case of error (search failure)
  - This caused noticeable (1-3 s) slowdowns when running tests targeting widgets with lots of children/descendants, such as the main window
- Add basic tests targeting navigating the tabs in the main interface


lbianchi-lbl:
After getting the GUI tests for #995 to work, I have a better understanding of the failures, and I believe they are caused by code running on threads other than the GUI thread accessing objects managed by Qt.

Many of these cases occur because of workarounds needed in the GUI testing framework to handle modal windows, so I'm working on implementing a better solution for that that doesn't rely on the thread-unsafe implementation. For this reason, we can close this and merge the relevant changes once in a new PR.

Issue #950, opened 2021-08-25
SDoE: Enhancements to SDoE module


sotorrio1:
- [x] Move **View** button from the **Output Directory** line to the first line of the design construction box called **Candidate File**.
- [x] Save the ID column when saving the designs created to a csv file

PR #961, opened 2021-09-16
SDoE: Adding couple of enhancements

sotorrio1:
Takes care of issue #950


sotorrio1:
@mirandamartin I've added the index column to USF and NUSF created designs. I haven't done for IRSF because as of now it doesn't make sense. The way the IRSF code provided by Towfiq works, it doesn't carry the index of the original candidate so this will need a major code re-factoring to make it work that way.

mirandamartin:
@sotorrio1 This version of FOQUS errored out on me ... sort of. I ran an NUSF design with the irsf-example.csv candidate set in the Examples folder. I chose 5 random starts as it said 30 seconds to run. Everything was working as expected. Including the random start choice with estimating the time it takes. Then once I clicked Run SDOE, I got a spinning wheel for at least ten minutes with nothing happening. I clicked StopSDOE eventually and it's like it kicked it into action and ran for a few seconds and gave me a design. The design doesn't look very good either. Attaching my terminal window. And the design scatterplots. 

Screen Shot 2021-09-17 at 6 07 51 PM

Weirddesign-091721

sotorrio1:
@mirandamartin I tried to reproduce your error without luck. I loaded the irsf_example.csv from the examples/tutorial_files/SDOE directory. Selected NUSF, opened the SDOE Dialog, Design size = 10, MWR = 5. Estimated run time for 10 random starts was 33 seconds so I ran that. I generated a design in the estimated time. Please see plots below. I can also send the generated design if needed.

SDOE_(NUSF)_Weight_Visualization_for_MWR=5

SDOE_Candidates_Visualization

sotorrio1:
@mirandamartin In fact, there shouldn't be any issues with the design generation since I haven't touched that code. All I did was moved the "View" button in the aggregated section in the main SDoE screen and allowed the automatically created index "__id" to be carried over into the visualization and final design.

mirandamartin:
@sotorrio1 I'm not sure what's going on...maybe I downloaded your branch wrong? I reran the example and got the same error, with it "running" for awhile (giving me the same error in terminal) until I clicked "Stop SDOE", then it ran as expected and produced a design in the expected amount of time. However, besides this issue, the fixes you've implemented I don't see. The view button is on the same line as it was before, and the index column did not appear in the csv file of the design. I'm going to ask Abby to try testing this today. In addition, I can delete the zip file and reinstall but because of proxy issues with anaconda I'm having, I have to wait until this evening when I'm home.

sotorrio1:
@mirandamartin If you're not seeing those two minor changes then you're either in the wrong branch or you didn't clone it right... Are you cloning my fork and then checking out the SDoE_enhancements branch?

mirandamartin:
@sotorrio1 I clicked the blue sotorrio1:sdoeEnhancements link at the top and then under the green CODE button selected to download the zip file. That's the normal way I view your branches. I made a new conda environment and installed foqus from the location of the zip file. Is this not right? I never actually learned how to look at branches another way so that's why I still do it this (probably not as simple) way.

sotorrio1:
@mirandamartin that way should also work... As you said, just make sure you deleted other versions and you start from a fresh environment :)

mirandamartin:
@sotorrio1 Cloning the repo and checking out 'SDoE_enhancements' worked! And the changes were there and look good.

mirandamartin (approved):
Looks good to me

PR #963, opened 2021-09-21
Address formatting issues in Turbine Settings message boxes

### LINKED ISSUE #958: UI Settings/Turbine Start Service PopUp String Format
# Description
After starting Turbine the information window that pops up contains text that is incorrectly formatted.
Screen Shot 2021-09-07 at 1 13 54 PM

lbianchi-lbl:
## Fixes #958 

- The formatting issues described in #958 are due to the fact that, by default, `subprocess.Popen.communicate()` returns `bytes` objects instead of `string`
- Passing `universal_newlines=True` when creating the `Popen` object ensures that (platform-independent) strings are returned instead
- `text=True` is a more readable alias, but this is not available for Python 3.6


lbianchi-lbl:
Also @ksbeattie this is unrelated from the PR itself, but something to think about: one of the advantages of running pylint separately on each supported Python version (as it is done in FOQUS) allows us to catch cases where some APIs have been changed in the standard library (such as `text=True` added in Python 3.7): https://github.com/CCSI-Toolset/FOQUS/pull/963/checks?sha=d141c03c90e6a6e7dec821849f8a15fc90f45e45

ksbeattie (approved):
I'm trusting that the review of it working on Windows from @lbianchi-lbl is enough!

PR #975, opened 2021-11-02
Use Black for code formatting

lbianchi-lbl:
In addition to the 200-odd-files diff, this PR includes:

- All modifications needed to restore the pylint directives that were deactivated after being shuffled around by Black during reformatting
- Add Black to `requirements-dev.txt`
- A GitHub Actions check that ensures that any committed code is Black-ified (or Black-compatible)
- A snippet to add to the developer docs (wiki, Sphinx, ...?) with a brief overview of how this works:

    ### Formatting code with Black

    Black (https://black.readthedocs.io) is the Python code formatter used by FOQUS to ensure that the codebase is formatted consistently.

    #### Installation

    Black is part of the FOQUS developer dependencies and it is installed alongside the other dependencies when running `pip install -r requirements-dev.txt` in the environment used for development.

    To verify that Black is installed correctly, run e.g.:

    ```sh
    black --version
    ```

    #### Usage

    Before committing code to the FOQUS repository, run `black` locally (i.e. on your development machine) with the default options from the repository root:

    ```sh
    black .
    ```

    #### Automated check

    When a PR is created or updated, as part of the automated check suite running on GitHub, Black will run in "check mode": rather than formatting files, it will check if any file would need to be formatted, and exiting with an error if this is the case.

    To reproduce this check locally (e.g. to verify that your code would pass this check when enforced on the GitHub side), from the root of the repository, run:

    ```sh
    black --check .
    ```


ksbeattie (approved):
Full steam ahead!

PR #983, opened 2021-11-22
Fixing link to contributors

ksbeattie:
The link was wrong.

(This might need to be done for all the other products in the CCSI-Toolset org...)


sotorrio1 (approved):
LGTM

Issue #969, opened 2021-10-05
SDoE: NUSF Error


sotorrio1:
When trying this in **NUSF**, I received weird errors upon selecting **Estimate Runtime**, which seem to be dependent on the design size I selected. I have no idea why NUSF is giving me errors and USF and IRSF ran fine with the exact same datasets. 

Clicking Estimate Runtime for design of size 6 (all else defaults) in **NUSF**:
<img width="955" alt="foqus-cands+prevdata-errorEstRuntime-nusf-size6" src="https://user-images.githubusercontent.com/55799605/135925895-c288f022-5313-434e-9a5e-86be63ddf4fb.png">

Just changed to size 5:
<img width="954" alt="foqus-cands+prevdata-errorEstRuntime-nusf-size5" src="https://user-images.githubusercontent.com/55799605/135925942-57b17bd9-b55c-4aa6-bee2-ac6ceb9404d4.png">

ksbeattie:
@sotorrio1 is working with the LANL folks on this.

jmorgan29:
I was working through an example for NUSF and found an error that seems similar to this, with slightly different syntax:
![NUSF error](https://user-images.githubusercontent.com/11306727/140406532-353a6e86-4e56-4256-941a-880e05be5f11.png)

I also found that the last line generally takes the form 'could not broadcast input array from shape(m,) into shape(m+n,) where m is the chosen design size and n is the number of observations in the 'previous data'. If no previous data are used for the NUSF design, this error does not show up when estimating the run time, and the design is generated properly.

ksbeattie:
@sotorrio1 is going to work with @pn51 on this, given his past experience with the NUSF numpy-ificaiton.

ksbeattie:
@sotorrio1, any progress on this?

sotorrio1:
@ksbeattie I started working on it with @pn51 and we're meeting again this Thursday. Hopefully we get to a solution to this issue then.

ksbeattie:
@sotorrio1, any new on this?  I think we were hoping to get a PR addressing this into the 3.12.0 release this week?

PR #992, opened 2021-12-20
Fix NUSF with history

pn51:
Addresses issue #969.

### DIFF
--- a/foqus_lib/framework/sdoe/nusf.py
+++ b/foqus_lib/framework/sdoe/nusf.py
@@ -20,8 +20,7 @@ import time
 import pandas as pd  # only used for the final output of criterion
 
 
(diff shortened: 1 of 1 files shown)


ksbeattie:
Hey @pn51, you should be able to fix the "black" failure by just running the command `black` on the one file you modified, then commit/push that change.

mirandamartin:
@pn51 Phan, would you have time tomorrow or Wednesday to walk me through your changes? I'd like to understand more of what you and Pedro did to fix the code.

sotorrio1 (approved):
Just tested running NUSF with previous data file and everything works as expected.

mirandamartin (approved):
Approved, though I couldn't tell much of a difference in designs before the most recent fix and after. We should reexamine in the future.

sotorrio1 (approved):
LGTM

PR #995, opened 2022-01-04
Machine Learning & Artificial Intelligence Plugin

bpaul4:
Building off FOQUS's existing plugin to load custom Pymodel scripts, this tool allows users to import Tensorflow Keras models into FOQUS nodes. As a first draft, the tool enables new flowsheet capabilities using the same UI as the Pymodel plugin:

- Users can save Sequential or Functional API Keras models as H5 files, and import into FOQUS nodes for model prediction. The tool will create default variable labels and bounds as needed
- Users can follow the documentation workflow and examples to create a CustomLayer class, and the plugin will extract any custom labels/bounds into the Node Editor
- Users can specify whether the training utilized data normalization, and the tool will handle needed scaling/unscaling internally (only linear scaling is currently supported)


bpaul4:
@lbianchi-lbl I have addressed some of the test failures, and the only remaining pylint/formatting issue is importing Tensorflow. The documentation builds successfully in my local environment and loads correctly on GitHub, and the cause of the Build docs failure to load certain PNG files is a bit unclear.

bpaul4:
@lbianchi-lbl your tests and CI updates are very well written, and it appears to have resolved all remaining GitHub Actions issues - thank you for your continuous efforts on this task.

@anujad95 @jmorgan29 @MAZamarripa please provide any additional feedback you have on the Plugin code or the new tests so we may merge these updates soon; @anujad95 and I previously tested the tool on our local machines with no issues.

foqus_lib/framework/graph/node.py
```diff
@@ -24,24 +24,40 @@
...
+from PyQt5.QtWidgets import QMessageBox
+from importlib import import_module
+
+# pylint: disable=import-error
+try:
+    # tensorflow should be installed, but is not required for non ML/AI models
+    import tensorflow as tf
+
+    load = tf.keras.models.load_model
+except:
```
lbianchi-lbl, line 48:
```suggestion
except ImportError:
    def load(*args, **kwargs):
        raise RuntimeError(f"`load()` was called with args={args}, kwargs={kwargs} but `tensorflow` is not available")
```

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+    def __init__(self, model):
+        pymodel.__init__(self)
+
+        # attempt to retrieve required information from loaded model, and set defaults otherwise
+        self.model = model
+
+        for i in range(np.shape(self.model.inputs[0])[1]):
+            try:
+                input_label = self.model.layers[1].input_labels[i]
+            except:
```
lbianchi-lbl, line 109: Suggestions to make exceptions more specific:

- Use `except MyExceptionType:` instead of "bare `except`
- Add a comment with more details on when that exception might occur (e.g. "this happens if the input labels in the model do not match those stored in the file" (just an example))

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+                input_label = "x" + str(i + 1)
+            try:
+                input_min = self.model.layers[1].input_bounds[input_label][0]
+            except:
+                input_min = 0  # not necessarily a good default
+            try:
+                input_max = self.model.layers[1].input_bounds[input_label][1]
+            except:
```
lbianchi-lbl, line 117: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+                vst="pymodel",
+                vdesc="input var " + str(i + 1),
+                tags=[],
+                dtype=float,
+            )
+
+        for j in range(np.shape(self.model.outputs[0])[1]):
+            try:
+                output_label = self.model.layers[1].output_labels[j]
+            except:
```
lbianchi-lbl, line 135: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+                output_label = "z" + str(j + 1)
+            try:
+                output_min = self.model.layers[1].output_bounds[output_label][0]
+            except:
```
lbianchi-lbl, line 139: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+                output_min = 0  # not necessarily a good default
+            try:
+                output_max = self.model.layers[1].output_bounds[output_label][1]
+            except:
```
lbianchi-lbl, line 143: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -80,6 +96,97 @@ def setCodeStrings(self):
...
+                vst="pymodel",
+                vdesc="output var " + str(j + 1),
+                tags=[],
+                dtype=float,
+            )
+
+        # check if user passed a model for normalized data - FOQUS will automatically scale/un-scale
+        try:  # if attribute exists, user has specified a model form
+            self.normalized = self.model.layers[1].normalized
+        except:  # otherwise user did not pass a normalized model
```
lbianchi-lbl, line 161: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -485,6 +592,26 @@ def setSim(self, newType=None, newModel=None, force=False, ids=None):
...
+            os.chdir(os.path.join(os.getcwd(), "user_ml_ai_models"))
+            try:  # see if custom layer script exists
+                module = import_module(str(self.modelName))  # contains CustomLayer
+                self.model = load(
+                    str(self.modelName) + ".h5",
+                    custom_objects={
+                        str(self.modelName): getattr(module, str(self.modelName))
+                    },
+                )
+            except:  # try to load model without custom layer
```
lbianchi-lbl, line 607: TODO: make `except` clause more specific (see above for suggestions)

foqus_lib/framework/graph/node.py
```diff
@@ -485,6 +592,26 @@ def setSim(self, newType=None, newModel=None, force=False, ids=None):
...
+                self.model = load(str(self.modelName) + ".h5")
+            os.chdir(cwd)  # reset to original working directory
```
lbianchi-lbl, line 609: Resetting the working dir inside a `finally` block ensures that it is always restored, regardless of what exceptions are thrown in the previous lines.

```suggestion
            finally:
                os.chdir(cwd)  # reset to original working directory
```

foqus_lib/framework/graph/node.py
```diff
@@ -959,3 +1088,36 @@ def killTurbineSession(self):
...
```
lbianchi-lbl, line 1111: See above for exception and `os.chdir()` handling

lbianchi-lbl (changes requested):
Following the discussion today during the dev call, we came to the conclusion that making the exception handling more specific could be a low-effort way to make unexpected errors more visible. I've left review comments where I think it would make sense to do this, as well as a few suggestions on the implementation.

foqus_lib/framework/graph/node.py
```diff
@@ -45,9 +45,12 @@
...
 
     load = tf.keras.models.load_model
-except:
-    pass  # errors will be thrown if tensorflow is called but not installed,
-    #  otherwise no error should be thrown so passing is fine
+except ImportError:
+    # if TensorFlow is not available, create a proxy function that will raise
+    # an exception whenever code tries to use `load()` at runtime
+    def load(*args, **kwargs):
+        raise RuntimeError(f"`load()` was called with args={args},"
```
lbianchi-lbl: Minor suggestion, but in retrospective `ModuleNotFoundError` would be a better fit than `RuntimeError` here.

foqus_lib/framework/graph/node.py
```diff
@@ -106,15 +109,33 @@ def __init__(self, model):
         for i in range(np.shape(self.model.inputs[0])[1]):
             try:
                 input_label = self.model.layers[1].input_labels[i]
-            except:
+            except AttributeError:
+                logging.getLogger("foqus." + __name__).info(
```
lbianchi-lbl, line 117: I think these extended messages are great, they add lots of useful context both for users and other developers.

One minor suggestion: instead of repeating the `logging.getLogger("foqus." + __name__)` each time, you could create a logger object at the module level, e.g. `_logger = logging.getLogger("foqus." + __name__)` after the `import` statements on Line 40, and then use that object throughout the rest of the module, i.e. `_logger.info(...)`

lbianchi-lbl (commented):
Thanks @bpaul4 for adding the more specific exceptions and especially the detailed messages, I think they're really helpful.

I've left two very minor corrections, and they're completely "take-it-or-leave-it", so feel free to dismiss them and I'm happy to approve either way.

lbianchi-lbl (approved):

bpaul4:
Thanks @lbianchi-lbl, I agree with your final two suggestions. I'll make those changes and push again for final approval.

lbianchi-lbl (approved):
Looks good!

lbianchi-lbl:
I believe that the last round of reviews and changes covers what we discussed on Tuesday, so I'm going ahead and merge this in. Thanks to all those involved and @bpaul4 in particular for bringing this over the finish line!

Issue #993, opened 2022-01-03
SDoE Usability Suggestion: Remove data imputation fields when there are no missing values


sspoon:
When loading a candidate set, the table describing the loaded files contains a workflow related to data imputation for missing values. If there are no missing values, these fields appear as inactive fields. If data imputation is needed, then the fields become active. See the inactive fields in the red boxes below:
![imputation](https://user-images.githubusercontent.com/5075865/147966680-73a461a4-dd64-428b-84b6-351a651b50fd.png)

![imputation2](https://user-images.githubusercontent.com/5075865/147966713-1eddad6f-28b9-428d-9d3e-0f448ae57334.png)

I would suggest instead to only show these fields if data imputation is needed, instead of having these fields inactive. It's unclear at this point how often the data imputation feature is needed, and having these fields inactive by default makes users feel that they have incomplete tasks at hand. So the default use case (no data imputation) would look like this:

![1U_starting_screen](https://user-images.githubusercontent.com/5075865/147966907-8b61ccd5-b169-4068-968a-242b643eddd5.png)

Then the fields can be added if data imputation is needed. 

If it becomes the case that data imputation is a common use case, I would suggest removing the workflow from the table entirely and moving it into a dialog window. But I think we should address this only if it becomes a more common use case.

ksbeattie:
@sspoon, could you review that @sotorrio1 fix is correct?

sspoon:
@ksbeattie @sotorrio1 The fix looks good!

PR #996, opened 2022-01-06
SDoE: Show/hide table columns dynamically

sotorrio1:
Depending on the existence of missing values in the uploaded candidate set, the columns in the **Design Setup** table will show/hide dynamically to avoid user confusion. Should take care of issue #993, suggested by @sspoon


mirandamartin (approved):
Looks good to me!

Issue #999, opened 2022-01-18
FOQUS Cloud: OUU bugs


boverhof:
# Description
When Running OUU via the FOQUS UQ, after hitting start an exception is happening before submission of jobs.

```bash
2022-01-18 11:03:47,752 - CRITICAL - foqus.foqus_lib.foqus - unhandled exception
Traceback (most recent call last):
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 570, in simSelected
    self.initUQToolBox()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1150, in initUQToolBox
    self.initData()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1177, in initData
    self.initFilter()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1190, in initFilter
    self.refreshFilterData()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1236, in refreshFilterData
    self.filterWidget.refreshContents()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/flowsheet/dataBrowserFrame.py", line 165, in refreshContents
    if self.results is None or self.results.empty:
AttributeError: 'SampleData' object has no attribute 'empty'
2022-01-18 11:04:31,206 - CRITICAL - foqus.foqus_lib.foqus - unhandled exception
Traceback (most recent call last):
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 570, in simSelected
    self.initUQToolBox()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1150, in initUQToolBox
    self.initData()
  File "/Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof/foqus_lib/gui/uq/uqSetupFrame.py", line 1177, in initData
    self.initFilter()
[...]

PR #1001, opened 2022-01-18
FOQUS Cloud: fixing OUU bug

sotorrio1:
Trying to fix issue #999


lbianchi-lbl:
The fix should work and we'd like to have it in, but it'd be nice if we could have a test for this. I'll check if I can quickly implement a test that can reproduce the error, and if not, we should just merge this fix.

lbianchi-lbl (approved):
LGTM

PR #1007, opened 2022-02-02
Create PULL_REQUEST_TEMPLATE.md

ksbeattie:
We should have one of these.


sotorrio1 (approved):
Interesting. How would this work?  We use the template in the PR body?

lbianchi-lbl (approved):
LGTM

PR #1034, opened 2022-06-03
ML AI Plugin Data Normalization Options

bpaul4:
## Fixes/Addresses:

- In addition to new features described below, this pull request addresses a gap in documentation and testing for the ML AI Plugin; additionally, optional dependency calls are now appropriately handled when not installed, and cited properly in the references.

## Summary/Motivation:

- The new scaling methods adjust handle user-trained neural networks that expect scaled data between [0, 1] using f(minimum value) = 0 and f(maximum value = 1); users must normalize and train Keras models with the correct attribute flag before passing to FOQUS.
- FOQUS handles SymPy similarly to TensorFlow in the test suite. Users can pass a string following SymPy syntax, and the plugin will internally build functions to scale/unscale variables when needed.
- The test module, documentation and examples underwent extensive expansion to cover the new capabilities.

- Support for new data normalization types (previously linear only, now log, power, log 2, power 2)
- Support for custom normalization functions with internal parsing via SymPy, a new optional dependency

### Legal Acknowledgement


bpaul4:
@lbianchi-lbl the latest version of the test module defines a single set of fixtures for as many objects as possible, launches a single main window instance and sequentially loads/runs the three example models. It runs well on all platforms and we may be able to implement similar "condensing" of repetitive tests to mitigate against other OS memory issues.

I believe the best way to increase the coverage is to write tests for `node.py` functionality and will use `nodeVars_test.py` as a guide. The documentation issue is resolved in #1035 .

MAZamarripa (commented):
lgtm

lbianchi-lbl (approved):
As far as the testing and CI implementation is concerned, this looks good. We might have to revisit some low-level details once the next batch of GUI testing fixes are in, but we should go ahead and merge this in the meantime.

PR #1040, opened 2022-06-28
Sdoe plot update

pn51:
## Summary/Motivation:
SDoE scatterplot enhancements to improve usability and assessment of design coverage/properties.

-Plots (N)USF candidate, design, and previous points and in different styles.
-Reweights transparency of candidate points by a specified weight.

### Legal Acknowledgement

### DIFF
--- a/foqus_lib/framework/sdoe/nusf.py
+++ b/foqus_lib/framework/sdoe/nusf.py
@@ -92,7 +92,7 @@ def update_min_dist(rcand, cand, ncand, xcols, wcol, md, mdpts, mties, dmat, his
             else np.concatenate((rcand[:, wcol], hist[:, wcol]))
         )
         m = row[xcols] - xs
-        m = np.sum(m ** 2, axis=1)
+        m = np.sum(m**2, axis=1)
         m = m * row[wcol] * weights
 
         dmat = np.copy(dmat_)
--- a/foqus_lib/framework/sdoe/nusf.py
+++ b/foqus_lib/framework/sdoe/nusf.py
@@ -92,7 +92,7 @@ def update_min_dist(rcand, cand, ncand, xcols, wcol, md, mdpts, mties, dmat, his
             else np.concatenate((rcand[:, wcol], hist[:, wcol]))
         )
         m = row[xcols] - xs
-        m = np.sum(m**2, axis=1)
+        m = np.sum(m ** 2, axis=1)
         m = m * row[wcol] * weights
 
         dmat = np.copy(dmat_)
--- a/foqus_lib/framework/sdoe/nusf.py
+++ b/foqus_lib/framework/sdoe/nusf.py
@@ -231,7 +231,8 @@ def criterion(
 ):
 
     ncand = len(cand)
-    nhist = len(hist)
+    if hist is not None:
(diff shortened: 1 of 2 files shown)


mirandamartin:
@pn51 When I tried to run an NUSF design, it gave me this error
2022-06-27 21:05:21,434 - CRITICAL - foqus.foqus_lib.foqus - unhandled exception
Traceback (most recent call last):
  File "/Users/mgmartin/Desktop/FOQUS-Other/FOQUS-phan-1/FOQUS/foqus_lib/gui/sdoe/sdoeAnalysisDialog.py", line 868, in testSdoeNUSF
    self.writeConfigFile(test=True), self.designSize_spin.value(), test=True
  File "/Users/mgmartin/Desktop/FOQUS-Other/FOQUS-phan-1/FOQUS/foqus_lib/framework/sdoe/sdoe.py", line 165, in run
    _results = criterion(cand, args, nr, nd, mode=mode, hist=hist)
  File "/Users/mgmartin/Desktop/FOQUS-Other/FOQUS-phan-1/FOQUS/foqus_lib/framework/sdoe/nusf.py", line 234, in criterion
    nhist = len(hist)
TypeError: object of type 'NoneType' has no len()

I tried two different candidate sets and got the same error. This was for NUSF designs.

mirandamartin:
@pn51 USF designs worked for me. However, I didn't see the candidate set plotted along with the design. 
sdoe-foqus-enhancements-usftesting

mirandamartin:
@pn51 I definitely see the candidate set now, thanks. It sounds like the NUSF issue may be a deeper problem we should revisit when you have time. However, I ran an NUSF design with previous data and it worked. In that case, the plots also look great. 

A few small points if these changes are quick to implement. Otherwise, we can go with what we have for now and come back to those things. 

1. The shading of the previous data / history looks great, except it makes more sense from a design sense to have the previous data points not show weights. They should just be solid points as the design points are. 
2. The candidate point shading doesn't seem to actually follow the weights. When I plot using the weight variable as a variable in the plot, the shading doesn't follow what the weights should be. For example, in the NUSF plots below, the top right plot should have shading going from lightest on the left to darkest on the right. Y is the weight and Y increases from left to right. But the weights are kind of all over the place. Surprisingly, the weight shading we currently see for previous data does look right. Can that version be implemented for the candidate set points?
3. For USF, I'm now seeing different shades of gray for candidate points. There are no weights for USF, can we make them all the same shade of gray?

NUSF design with previous data. Y is weighting variable, X1 and X2 are inputs.
2022-06-28-enhancements-nusf-prevdata

USF design. 
2022-06-28-enhancements-usf

mirandamartin:
Sorry, I attached out of date screenshots. These are from after the last commit. 
NUSF design with previous data. Weights are the Y variable and should be getting darker left to right on those right side scatterplots.
2022-06-28_enhancements_nusfprevdata_updated

USF design. Not sure why we have different colors for candidate points. There are no weights in USF so all should be the same color gray. Otherwise looks great!
2022-06-28_enhancements_usf_updated

mirandamartin (approved):
It looks so good, thanks so much for all your work on this!

PR #1045, opened 2022-07-18
Core Node Functionality Tests

bpaul4:
## Fixes/Addresses:

- Addresses code coverage gap concerning core FOQUS node functionality
- Modifications to existing ML AI class tests
- New tests covering key node methods for model types None, Pymodel , ML AI, and Turbine

-
-

### Legal Acknowledgement


bpaul4:
There are two outstanding items I would appreciate feedback on:

1. @lbianchi-lbl , I commented out the test deserializing the Keras classes as it seems the registration is now temporary and is reset for the later tests. Including the deserialization actually broke later tests expecting to find those classes. I wanted to get your opinion on this behavior before deleting it completely.
2. @boverhof I dived a bit deeper into the Turbine configuration issue, and found that the `response` from `urllib.request.urlopen` is returning an `_io.BytesIO` object instead of an HTTP request object. I'm working to increase my understanding of the Turbine server and would appreciate your thoughts on this.

lbianchi-lbl:
@bpaul4 For point 2: I can confirm I can reproduce the failures locally. I've looked into the case of the `urllib.request.urlopen()` errors, and I think I've tracked it down to this test file:

https://github.com/CCSI-Toolset/FOQUS/blob/eb8586d36150d601615a2c61ff2331a3917a04d6/foqus_lib/service/test/flowsheet_control_test.py#L50-L54

My guess is that the monkey-patching (overriding predefined functions at runtime) that is performed for the `test_floqus_aws_config()` test function is *not* undone at the end of the test, and therefore is still in its mocked (and therefore, now invalid) state when it is called again for the node tests.

I think this is something we should address in `flowsheet_control_test.py`, to enforce the "campsite principle" for tests ("always clean up after yourself"). However, I personally would want to understand that code a bit better before changing it, so in the meantime, an easier solution would be to remove any test that's creating `TurbineConfiguration` instances.

For point 1: I'm not sure I understand the issue. Could you provide some clarification?

bpaul4:
1. The tests are now referencing the session-scoped temporary directory for model files, and "popping" the registered Keras classes means later tests fail to find the classes. I think changing to a temporary directory before the ML AI node tests affects the registration. Since the deserialization's purpose was solely to allow tests to run, I am fine deleting it and wanted to check with you before doing so.
2. I support removing the two Turbine tests for now while we investigate `flowsheet_control_test.py`, and merging this PR without them.

lbianchi-lbl (approved):
Looks good :+1:

PR #1046, opened 2022-07-19
Resolve crashes in GUI tests

lbianchi-lbl:
## 

(replace this with the issue # fixed or addressed, if no issue exists then a brief statement of what this PR does)

## Summary/Motivation:

- Addresses intermittent failures for the GUI tests due to segfaults

- Change scope for several fixtures (most importantly `main_window` to `session` scope)

### Legal Acknowledgement


foqus_lib/gui/ouu/ouuSetupFrame.py
```diff
@@ -680,7 +680,7 @@ def clearPlots(self):
         self.objXPoints = []
         self.objYPoints = []
         if "objFigAx" in self.__dict__ and len(self.objFigAx.lines) > 0:
-            self.objFigAx.lines = []
+            self.objFigAx.lines.clear()
```
bpaul4, line 683: Seems reasonable, curious if this is related to the rest of this PR or just something you spotted.

bpaul4 (approved):
Looks very nice, great job @lbianchi-lbl and thanks for doing this. I have a couple comments/suggestions that are not critical.

foqus_lib/gui/ouu/ouuSetupFrame.py
lbianchi-lbl: Good question, yeah, this is something that I believe is completely unrelated, but it did cause an error (`Can't set attribute`) if left as it was. My best guess is that at some point the `lines` attribute of that object (whatever is returned by `matplotlib.Figure.make_subplots()`) was turned into a read-only [property](https://docs.python.org/3/library/functions.html#property).

foqus_lib/conftest.py
```diff
@@ -49,22 +56,56 @@ def psuade_path():
...
-    return str(examples_dir / "test_files" / request.param)
+@pytest.fixture(
+    scope="session",
+)
+def foqus_working_dir(
+    request, tmp_path_factory, name: str = "foqus-working-dir"
+) -> Path:
+    d = tmp_path_factory.mktemp(name, numbered=False)
+
+    what = "The FOQUS working directory for this test run"
```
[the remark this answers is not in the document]
lbianchi-lbl: Yeah, I agree that the naming for this is not the clearest (and I admit this would be far from being the worst offender within the `conftest.py`/`pytest_qt_extras.py` code...). Actually, I've since added better reporting functionality for displaying the working directory (that, unlike `print()` or `logging`, is also shown when pytest is run without the `-s` flag), so I think we can get rid of those printouts entirely.

bpaul4 (approved):

PR #1049, opened 2022-08-09
Support For Additional ML AI File Types

bpaul4:
- Allows users to save Keras models as SavedModel file format in addition to H5
- Adds similar structure to support JSON model files
- Updated example files, tests and documentation

### Legal Acknowledgement


bpaul4:
I believe the py3.8/win64 error is due to auto-normalizing line endings - in the commits above I found that applying `.gitattribute` to all `.pb` files makes `saved_model.pb` unparseable, as well as the fact that the flags used to check-in the files during `git add` seem to hugely impact how they behave once they reach the remote branch. This error stopped appearing locally after regenerating the file, so I think it's being altered before it reaches the CI run on GitHub Actions. If there is a way to ensure that git does not alter the file at all while checking-in or running the tests (copying to a new temp directory), that should resolve this issue.

Daison2102 (approved):
Tested this and it looks good!

bpaul4:
Thanks @Daison2102 for testing the changes. The current failures are related to the UQ tests, and I believe they are not related to the changes here (I'm investigating the cause).

lbianchi-lbl (approved):

PR #1069, opened 2022-10-11
check for empty Node Vector Variables 

boverhof:
## Fixes/Addresses:

input/output vector variables for a NODE are initialized as empty dict and this is wrongly being compared and processed against the parent graph as missing.  Fix Test if it's empty dict, if it is don't process it.

-
-

### Legal Acknowledgement


ksbeattie (dismissed):
Act of faith

Issue #1074, opened 2022-10-14
Tests for plugin loading from string reading


bpaul4:
Recently, I noticed that FOQUS is not correctly loading Pymodel plugins from the user_plugins or examples/.../pymodels folders. A recent update had run `black` on `session.py` and `surrogates.py` which generate the idString objects used to detect and load plugins, so that files contained `black`-formatted `# FOQUS-[tag]` while the search expected `#FOQUS-[tag]`.

The issue went undetected until `black` was run on the working directory and those plugins failed to load; the console did not display any error or warning message that the idString did not properly parse. There should be tests that check that plugins are correctly loaded when FOQUS is launched, and we should add better logger messages for when this case arises.

lbianchi-lbl:
Hi @bpaul4, thanks for spotting this and creating an issue.

Can I ask you to point out where exactly the files for the plugins that are failing to load are located within the FOQUS repository? I'm starting to look into this (also in relation to #908) and I'd like to make sure I'm at least aware of the existing plugins.

PR #1075, opened 2022-10-17
Fixing Plugin Load Issue and Adding Plugin Tests

bpaul4:
## Fixes/Addresses:

Addresses #1074 by fixing idString objects and adding tests.

- Fix so that plugins load correctly
- First attempt at adding tests to catch when plugin functionality is not working

### Legal Acknowledgement

### DIFF
--- a/foqus_lib/conftest.py
+++ b/foqus_lib/conftest.py
@@ -49,6 +49,13 @@ def foqus_examples_dir(_repo_root):
 
 
 @pytest.fixture(scope="session")
+def foqus_pymodels_dir(_repo_root):
+    _pymodels_dir = _repo_root / "foqus_lib" / "framework" / "pymodel"
+    assert _pymodels_dir.exists()
+    return _pymodels_dir
+
+
+@pytest.fixture(scope="session")
 def psuade_path():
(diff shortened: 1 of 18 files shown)


lbianchi-lbl:
@bpaul4 I think the "BFGS" errors might be due to the fact that the "magic" string currently used to load plugins is not required to be in a comment. I've just checked and the plugin file where the BFGS solver implementation is defined looks like this: https://github.com/CCSI-Toolset/FOQUS/blob/1207c928e24b873ca51200c738b07d737a468700/foqus_lib/framework/optimizer/BFGS.py#L16

Since it's in a docstring, it was "spared" by the Black formatting run that introduced a space between the `#` and `FOQUS_XYZ_PLUGIN`. So, after modifying the "magic" string in this PR, that plugin is not being picked up anymore which in turn causes the tests that expect it to be present to fail.

bpaul4:
Thanks @lbianchi-lbl for tracking down the root cause. The tests should pass once #1082 is merged.

ksbeattie (dismissed):
LGTM

lbianchi-lbl (approved):

PR #1082, opened 2022-10-26
Update `pytest-qt` to resolve `ModuleNotFoundError:`

bpaul4:
- Recently, `pytest-qt` removed support for the package `py` which broke our CI runs
- This re-pins the package from 4.0.* to 4.2.* to resolve the issue

### Legal Acknowledgement

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,7 +5,7 @@
 pylint==2.14.1
 # also pinning the version for astroid (used by pylint to analyze the AST)
 # since there can be significant differences between (non-major) versions,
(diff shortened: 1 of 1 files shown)


lbianchi-lbl (approved):
Thanks @bpaul4 for noticing this and coming up with a fix. I've looked into it myself, and I can confirm that this was due to an incompatibility between `pytest-qt` (which was pinned to 4.0.*) and `pytest` 7.2: as `pytest` was not pinned, it got updated to the latest version that turned out to be incompatible with `pytest-qt<4.2`.

PR #1094, opened 2022-12-02
Support For Scikit-Learn Machine Learning Models

bpaul4:
- Demonstrates how users can train/save Scikit-Learn neural networks with custom objects and produce .pkl files compatible with the ML AI Plugin in FOQUS
- Support for Scikit-Learn models in FOQUS, including updates to the class structure, search method, example files, documentation, tests, and optional dependencies list

### Legal Acknowledgement


foqus_lib/framework/graph/node.py
```diff
@@ -143,10 +143,32 @@ def torch_float(*args, **kwargs):
     return torch_load, torch_tensor, torch_float
 
 
+def attempt_load_pickle(try_imports=True):
```
lbianchi-lbl, line 146: I've just double-checked to be sure, but `pickle` is part of the Python standard library, so I don't think we need to check and/or handle the case when it might not be installed (this applies here, as well elsewhere in the code).

For this reason, I'd recommend to remove the import-checking code before merging. However, since strictly speaking they're only unnecessary rather than incorrect, we could also consider keeping them in case their presence makes other parts of the code significantly simpler.

foqus_lib/gui/tests/test_ml_ai.py
```diff
@@ -196,6 +202,7 @@ def test_flowsheet_run_successful(
 
     def test_load_and_run_meacustomnormformjson(self, active_session, simnode):
         pytest.importorskip("tensorflow", reason="tensorflow not installed")
+        pytest.importorskip("json", reason="json not installed")
```
lbianchi-lbl, line 205: Similarly to `pickle`, we don't need to handle the case where `json` is not importable.

foqus_lib/unit_test/node_test.py
```diff
@@ -349,15 +373,34 @@ def example_6(self, model_files):  # custom layer with custom normalization form
...
-        assert os.path.exists(model_pt[0])
         # has a custom layer with a custom normalization option
         model = torch_load(str(model_pt[0]))
 
         return model
 
+    @pytest.fixture(scope="function")
+    def example_7(self, model_files):  # custom layer with custom normalization form
+        # no tests using this fixture should run if pickle and sklearn are not installed
+        pytest.importorskip("pickle", reason="pickle not installed")
```
lbianchi-lbl, line 384: See above for `pickle` import handling.

foqus_lib/unit_test/node_test.py
```diff
@@ -406,6 +449,15 @@ def test_build_and_run_as_expected_6(self, example_6):
         test_pymodel = pymodel_ml_ai(example_6, trainer="torch")
         test_pymodel.run()
 
+    def test_build_and_run_as_expected_7(self, example_7):
+        # only run if pickle and sklearn are available; test run for custom norm example
+        pytest.importorskip("pickle", reason="pickle not installed")
```
lbianchi-lbl, line 454: See above for `pickle` import handling.

foqus_lib/unit_test/node_test.py
```diff
@@ -1394,3 +1458,139 @@ def test_runPymodelMLAI_example5(self, node, model_files):
...
+                    v, attribute
+                )
+            for vkey, v in inst.outputs.items():
+                assert getattr(node.gr.output[node.name][vkey], attribute) == getattr(
+                    v, attribute
+                )
+
+    def test_setSim_modelMLAI_example7(self, node, model_files):
+        # skip this test if pickle is not available
+        pytest.importorskip("pickle", reason="pickle not installed")
```
lbianchi-lbl, line 1532: See above for `pickle` import handling.

.github/workflows/checks.yml
```diff
@@ -73,7 +73,7 @@ jobs:
           - python-version: "3.8"  # avoid uploading coverage for full matrix
             use_coverage: true
           - python-version: "3.8"  # this is to avoid installing optional dependencies in all environments
-            optional-dependencies: tensorflow sympy torch
+            optional-dependencies: tensorflow sympy torch sklearn
```
lbianchi-lbl, line 76: Here I believe `optional-dependencies` expects a package distribution name (i.e. what you `pip install`), rather than a package name (i.e. what you `import`), therefore I think it should be:

```suggestion
            optional-dependencies: tensorflow sympy torch scikit-learn
```

lbianchi-lbl (changes requested):
The code in general looks good (although I haven't tested the actual workflow). I've left a few comments for things that we should consider addressing before merging, but there are no major blockers.

.github/workflows/checks.yml
bpaul4: Thanks @lbianchi-lbl, I will address your comments and check that everything runs smoothly.

foqus_lib/framework/graph/node.py
bpaul4: I will change this to check for `sklearn` instead, since this is the only context in which `pickle_load()` is called, and update the test skip markers appropriately.

examples/other_files/ML_AI_Plugin/mea_column_model_training_customnormform_scikitlearn.py
```diff
@@ -0,0 +1,84 @@
+import numpy as np
+import pandas as pd
+import random as rn
```
Daison2102, line 3: Imported but never used. Consider remove it

examples/other_files/ML_AI_Plugin/mea_column_model_training_customnormform_scikitlearn.py
```diff
@@ -0,0 +1,84 @@
...
+from sklearn.neural_network import MLPRegressor
+import pickle
+from pathlib import Path
```
Daison2102, line 6: Imported but never used. Consider remove it

Daison2102 (approved):
Run successfully tests and examples for sklearn models. LGTM

Daison2102:
Run successfully tests and examples for sklearn models. LGTM

Daison2102 (approved):

Daison2102 (approved):

lbianchi-lbl (approved):

Issue #1107, opened 2023-02-15
Tests skipped by default?


boverhof:
# Description
all the tests are being skipped because tensorflow isn't installed.

```
(py3.9) joshuas-mbp:foqus-boverhof boverhof$ pytest -v
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.9.16, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/boverhof/Documents/workspace.ccsi2/py3.9/bin/python
cachedir: .pytest_cache
PyQt5 5.15.7 -- Qt runtime 5.15.2 -- Qt compiled 5.15.2
rootdir: /Users/boverhof/Documents/workspace.ccsi2/foqus-boverhof, configfile: pytest.ini, testpaths: foqus_lib
plugins: qt-4.0.2, typeguard-2.13.3, cov-3.0.0
collected 150 items                                                                                                                                         

foqus_lib/framework/sampleResults/test/results_s3_test.py::test_results_empty SKIPPED (tensorflow not installed)                                      [  0%]
foqus_lib/framework/sampleResults/test/results_s3_test.py::test_add_result_single SKIPPED (tensorflow not installed)                                  [  1%]
foqus_lib/framework/sampleResults/test/results_s3_test.py::test_results_session_broke_result_page SKIPPED (tensorflow not installed)                  [  2%]
foqus_lib/framework/sampleResults/test/results_s3_test.py::test_results_session_result_page_1 SKIPPED (tensorflow not installed)                      [  2%]
foqus_lib/framework/sampleResults/test/results_s3_test.py::test_results_session_result_page_1_to_9 SKIPPED (tensorflow not installed)                 [  3%]
foqus_lib/framework/sdoe/test/test_irsf.py::test_inv_scale_cand SKIPPED (tensorflow not installed)                                                    [  4%]
foqus_lib/framework/sdoe/test/test_order.py::test_mat2tuples SKIPPED (tensorflow not installed)                                                       [  4%]
foqus_lib/framework/sdoe/test/test_order.py::test_rank SKIPPED (tensorflow not installed)                                                             [  5%]
foqus_lib/gui/tests/test_main_window.py::test_main_window_opens SKIPPED (tensorflow not installed)                                                    [  6%]
foqus_lib/gui/tests/test_ml_ai.py::test_model_files_are_present SKIPPED (tensorflow not installed)                                                    [  6%]
[...]

PR #1108, opened 2023-02-15
Remove calls to pytest.importorskip() from session-scoped fixtures

lbianchi-lbl:
## Summary/Motivation

- [`pytest.importorskip("my_module")`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-importorskip) causes tests to be skipped if `my_module` can't be imported
- This is used in FOQUS to account for tests that require optional dependencies such as `tensorflow`
- However, because of a subtlety of its implementation (basically it raises a special exception), if `pytest.importorskip()` is used within a session-scoped fixture (as opposed to inside a test function, or at the top of a test module e.g. `test_my_optional_feature.py`), it will cause *all* tests to be skipped since session-scoped fixture functions are run before any other fixture or test

- Remove calls to `pytest.importorskip()` from session-scoped fixtures

## How to check that this works?

- In our GitHub Actions workflow, the optional dependencies are only installed for some of the OS/Python version job matrix combinations (namely, they're installed on Python 3.8, and not installed on Python 3.9, with the choice being essentially arbitrary)
- By selecting a Python 3.9 run (i.e. where the ML/AI optional dependencies are not available) we see that only some of the tests are skipped, as opposed to all tests being skipped as reported in #1107

image

### Legal Acknowledgement

### DIFF
--- a/foqus_lib/conftest.py
+++ b/foqus_lib/conftest.py
@@ -115,8 +115,6 @@ def install_plugin_model_files(
 def foqus_ml_ai_models_dir(
     foqus_working_dir: Path,
 ) -> Path:
-    # skip this test if tensorflow is not available
-    pytest.importorskip("tensorflow", reason="tensorflow not installed")
 
     return foqus_working_dir / "user_ml_ai_models"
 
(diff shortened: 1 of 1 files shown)


bpaul4 (approved):

ksbeattie (approved):
LGTM

Issue #1114, opened 2023-03-06
No Simulation Message


boverhof:
# Description
Need a better no simulation message.
- Current Message:  'The simulation completed with an error 20'
- Issue:  Didn't load the simulation into Turbine
- Need to look at logs in order to understand what happened.

- [ ] Change Message in pop up "Simulation {{ name }} is Missing, goto 'Add/upload model to Turbine under session "

## Steps to Reproduce
- Load up a flowsheet that contains a node with a model [BFB](https://github.com/CCSI-Toolset/FOQUS/blob/master/examples/test_files/OUU_Remote_Turbine_BFB/BFB_OUU_06_09_2020.foqus)
- Hit "Run"

```
Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\foqus\lib\site-packages\foqus_lib\framework\sim\turbineConfiguration.py", line 313, in getSimApplication
    raise TurbineInterfaceEx(
foqus_lib.framework.sim.turbineConfiguration.TurbineInterfaceEx: 360 - Could not find simulation, Could not find simulation: HighConcentration





During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\ProgramData\Anaconda3\envs\foqus\lib\threading.py", line 932, in _bootstrap_inner
    self.run()
  File "C:\ProgramData\Anaconda3\envs\foqus\lib\site-packages\foqus_lib\framework\graph\graph.py", line 882, in run
    _log.exception("Exception in graph thread: %s" % (self.ex))
TypeError: not all arguments converted during string formatting
2023-03-06 20:14:09,118 - ERROR - foqus.foqus_lib.gui.main.mainWindow - to results graph thread was likely terminated
```

<img width="975" alt="Screen Shot 2023-03-06 at 11 14 24 AM" src="https://user-images.githubusercontent.com/738295/223208358-22bb2ab9-118b-424b-b5ef-7f5075eb148c.png">

kbuma:
I did a clean install on MacOS -> ran foqus -> loaded the example you listed -> tried to run it

It looks like it tried to connect to a local turbine instance (which I don't have running as I am on Mac OS) and provided a different cryptic error.

![image](https://github.com/CCSI-Toolset/FOQUS/assets/2824957/43419c04-a0fb-49aa-a913-866d38e67621)

Traceback (most recent call last):
[...]

PR #1147, opened 2023-06-13
adding attribute check to avoid throwing exception

kbuma:
## Fixes/Addresses:

This exception handling code was throwing an exception because the attribute being checked was not present.

## Summary/Motivation:

When looking to reproduce issue #1114 ran into this code throwing an exception.

- checking to see if attribute is present on object

### Legal Acknowledgement


ksbeattie (approved):
LGTM

Issue #1100, opened 2023-01-13
FOQUS fails to install on Apple M1/ARM platforms


ksbeattie:
In particular, it's the PyQt5 that is the problem, there is no pip package for that architecture.  There is a PyQt6 package, but then FOQUS won't work with it.

Rosetta might be a work around for those on the M1/ARM chips.

kbuma:
This appears to be due to there not being a wheel for this architecture thus requiring it to be compiled from source. Compiling from source fails because:
1) If you do not have qmake in you path it cannot build the package. To remedy this you can install Qt5 and add qmake to your path (e.g. using homebrew)
2) The build process for PyQt 5 has an interactive prompt that requires you to accept the license. This does not interact well with pip and causes the process to hang. (if you run pip in verbose mode you will see this)

My solution was to download and install PyQt based on the "Building and Installing from Source" instructions from https://www.riverbankcomputing.com/static/Docs/PyQt5/installation.html. Once that was done I could follow the normal instructions for installing FOQUS.

kbuma:
@ksbeattie It looks like @bpaul4 has suggested a simpler solution in
https://github.com/CCSI-Toolset/FOQUS/discussions/1146
We should update our documentation with that.

I've verified that that works on my machine.

bpaul4:
@eslickj identified the fix, and I agree it would be useful to update the documentation. According to the installation page we officially support Python 3.7 to 3.9, and our ARM fix uses 3.10. We should update the documentation if we do support 3.10 now.

lbianchi-lbl:
This is great, thanks all for working on testing and documenting this.

Just adding a couple of items:

- [ ] Try if `python=3.10` is a requirement, or things also work with 3.8 and 3.9
- [ ] Run the complete test suite, including the GUI tests (@lbianchi-lbl to provide a command)

kbuma:
verified that instructions from https://github.com/CCSI-Toolset/FOQUS/discussions/1146 work with 3.8

lbianchi-lbl:
@kbuma I've created a wiki page on how to run GUI tests locally: https://github.com/CCSI-Toolset/FOQUS/wiki/howto-gui-tests. Let me know if you run into any errors so we can update those steps for future reference.

kbuma:
Looks like the tests passed.

[test-results.txt](https://github.com/CCSI-Toolset/FOQUS/files/11739827/test-results.txt)

[...]

PR #1148, opened 2023-06-13
Changes to allow install for arm64 darwin.

kbuma:
## Fixes/Addresses:

#1100

## Summary/Motivation:

Allows users of MacOs on Apple silicon to install and run FOQUS.

- update to setup.py to exclude PyQt5 install on these machines
- documentation of steps required to install on these machines (once we have the next release out we can remove the need to download source as part of this)

### Legal Acknowledgement

### DIFF
--- a/docs/source/chapt_install/index.rst
+++ b/docs/source/chapt_install/index.rst
@@ -9,6 +9,10 @@ optional components for use within FOQUS.
 Quick Start
 -----------
 
+.. note::
+    If you are installing on Apple silicon please use the 
+    sub-sections as this quick start will not work.
+
 For those familiar with the details, here is a summary of how to install and run
 FOQUS:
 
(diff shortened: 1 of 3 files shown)


lbianchi-lbl (approved):

kbuma:
@lbianchi-lbl it looks like the py3.9 tests are no longer passing and that's blocking this from being merged. 

I don't think it's due to this change: I ran the tests on my developer machine and I'm getting the same behavior for the 3.17.1 release. I don't know enough about the tests to be able to tell why. Maybe new versions of some dependencies?

lbianchi-lbl:
> @lbianchi-lbl it looks like the py3.9 tests are no longer passing and that's blocking this from being merged.
> 
> I don't think it's due to this change: I ran the tests on my developer machine and I'm getting the same behavior for the 3.17.1 release. I don't know enough about the tests to be able to tell why. Maybe new versions of some dependencies?

Yeah, my best guess is NumPy: #1151

PR #1153, opened 2023-06-27
Initial draft for experimental plugin implementation

lbianchi-lbl:
### Legal Acknowledgement


lbianchi-lbl:
This is just an example at the moment, but here's how it looks like if we add an explicit check for PSUADE in `sdoeSetupFrame.py`:

image

The FOQUS interface starts as usual, but the SDoE frame won't be loaded. Eventually we'd also want the corresponding icon in the toolbar to be grayed out and/or display an error message/icon, but that require further work to move the code that loads the `QAction` for each frame from their current location in `mainWindow.py` to each plugin's module.

Issue #1126, opened 2023-03-26
ODoE Tutorial Issue


jmorgan29:
I am running the tutorial "Example 1: ODoE with Existing Candidate Set" in FOQUS version 3.16.0. I am able to complete the tutorial up to Step 11, when clicking on the "Run ODoE" button results in a TypeError associated with the file \foqus_lib\gui\sdoe\sdoeSetupFrame.py

For some additional background, I am following the tutorial instructions of using G-opt as the optimality criterion with a design size of 2 and number of restarts chosen as 3. There is also an additional method option for 'Fisher' and 'Bayesian', which seems to have been added in a later version than that used for the tutorial documentation. I have tested both methods and got the same error.

![ODOE Error](https://user-images.githubusercontent.com/11306727/227803536-67b661ac-487d-44cb-bc33-dc437c311779.png)

lbianchi-lbl:
@jmorgan29 thanks for opening the issue. From a quick look, I think the problem could be due to how the PSUADE output is parsed in `foqus_lib.framework.sdoe.odoeu.odoeu()`:

https://github.com/CCSI-Toolset/FOQUS/blob/5588689cdd64bc3d7270d129bb7c3330f4a6e0a1/foqus_lib/framework/sdoe/odoeu.py#L229-L241

More specifically, if the regular expression at L232 doesn't match, `best_optval` is left to its default value of `None` which causes the error in `sdoeSetupFrame.runOdoe()` since the format string expects `best_optval` to be a number instead.

Maybe @sotorrio1 can chime with more details and/or provide some example PSUADE output files/strings that we could use to add tests to exercise the parsing functionality.

sotorrio1:
@jmorgan29 thanks for opening the issue, I'll definitely update the documentation to show the new method ("Fisher" vs "Bayesian") and the new optimality criterion E-Opt. 

I just ran tutorial one for ODoE and I can run it without being able to reproduce your error. What version of PSUADE are you using? You should make sure you [installed version 1.9](https://foqus.readthedocs.io/en/stable/chapt_install/install_optional.html#install-psuade-lite-current-version-1-9-0), since 1.9 will be the only version supporting the newest version of ODoE. I will add a note to the tutorials as well to avoid confusions. 

@lbianchi-lbl I'd love to help regarding adding some testing to the parsing functionality. We're  not writing to disk but extracting the information from the PSUADE console, but I can definitely provide inputs.

sotorrio1:
[...]

PR #1159, opened 2023-08-23
ODoE: Error handling when results are None

sotorrio1:
## Fixes/Addresses:

Addresses issue #1126 

## Summary/Motivation:
Although issue #1126 seems to be solved, we're proposing a different way to handle errors in the case the ODoE results (best_indices and best_optval) are None. 

- check if best_indices is None and modify resultMessage accordingly 
- check if best_optval is None and modify resultMessage accordingly 

### Legal Acknowledgement


ksbeattie (approved):
LGTM.  At least this doesn't throw an exception and shows the user that something has gone wrong.

PR #1183, opened 2023-12-14
Ensure paths mentioned in Sphinx docs are valid

lbianchi-lbl:
- Add small Sphinx extension adding `:path:` and `:filename:` roles to display paths relative to repository root with validation (a warning is emitted if the referred path is not found during the build)
- Convert a subset of the (non-role) paths found in the docs to use the `:path:`/`:filename:` roles
- Fix non-existing paths until build passes

This is how it looks when non-existing paths are used:

image

### Legal Acknowledgement

### DIFF
--- a/docs/source/_repopath.py
+++ b/docs/source/_repopath.py
@@ -24,18 +24,6 @@ def _get_repo_root() -> Path:
     return Path(repo_root).resolve()
 
 
-class RepoPath(EmphasizedLiteral):
-    @property
-    def path(self) -> Path:
-        repo_root = _get_repo_root()
-        return repo_root / self.text
-
-    def run(self):
-        if not self.path.exists():
-            _logger.warning("path %r does not exist", self.text)
-        return super().run()
-
-
 class PathRoleBase:
     def __init__(self, base_path=""):
         self.base_path = Path(base_path)
--- a/docs/source/_repopath.py
+++ b/docs/source/_repopath.py
@@ -61,6 +61,16 @@ class Relpath(PathRoleBase):
 
 
 class Filename(PathRoleBase):
+    def validate(self, path: Path) -> None:
+        if not path.is_file():
+            _logger.warning(
+                "No file named %r in directory %r", path.name, os.fspath(path.parent)
+            )
+        if not path.stat().st_size > 0:
+            _logger.warning(
+                "File %r in %r exists but it's empty", path.name, os.fspath(path.parent)
+            )
+
     def display(self, path: Path, text="") -> str:
         return path.name
 
--- a/docs/source/_repopath.py
+++ b/docs/source/_repopath.py
@@ -32,7 +32,9 @@ class PathRoleBase:
         if not path.exists():
             relpath = path.relative_to(self.base_path)
             _logger.warning(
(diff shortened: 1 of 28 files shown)


docs/source/_repopath.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def run(self):
+        if not self.path.exists():
+            _logger.warning("path %r does not exist", self.text)
+        return super().run()
+
+
+class PathRoleBase:
+    def __init__(self, base_path=""):
+        self.base_path = Path(base_path)
```
kbuma, line 41: why the Path constructor here? all the usage already passes a Path into the PathRoleBase

docs/source/_repopath.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def validate(self, path: Path) -> None:
+        if not path.exists():
+            relpath = path.relative_to(self.base_path)
+            _logger.warning(
+                "path %r does not exist within %r", str(relpath), str(self.base_path)
```
kbuma, line 47: for consistency maybe use os.fspath() rather than str() for both of these

docs/source/_repopath.py
```diff
@@ -0,0 +1,84 @@
...
+        node = nodes.literal(path_to_display, path_to_display)
+        return [node], []
+
+
+class Relpath(PathRoleBase):
+    def display(self, path: Path, text="") -> str:
+        return os.fspath(path.relative_to(self.base_path))
+
+
+class Filename(PathRoleBase):
```
kbuma, line 75: maybe override validate to add a is_file() check as well for this type

kbuma (dismissed):
excellent solution. added just a few nit-picks in _repopath.py

docs/source/_repopath.py
```diff
@@ -0,0 +1,84 @@
...
+    def display(self, path: Path, text="") -> str:
+        return path.name
```
kbuma, line 77: consider returning something like `path.name + ", and this file is located in " + os.fspath(path.parent.relative_to(self.base_path))` since that pattern occurs repeatedly in the usage

docs/source/_repopath.py
lbianchi-lbl: Good point, I think I was just being lazy :sweat_smile:

docs/source/_repopath.py
lbianchi-lbl: Mostly for [robustness](https://en.wikipedia.org/wiki/Robustness_principle)/defensiveness: the internals only work if `self.base_path` is a `pathlib.Path`, and `pathlib.Path` instances are immutable so creating a new `Path` is easier than having `isinstance(..., Path)` checks.

docs/source/_repopath.py
lbianchi-lbl: Good idea, I'll add that.

docs/source/_repopath.py
lbianchi-lbl: Good suggestion, maybe to keep in mind for a future iteration? I see the pattern you're referring to, but having the role handle more complicated formatting (since the prose would have to be formatted differently than the actual path) looked quite a bit trickier than the current implementation.

ksbeattie (approved):
LGTM

Issue #1111, opened 2023-02-21
Add type annotations to SDoE


franflame:
Add typing to SDoE functions for testing purposes, prioritize usf.py

lbianchi-lbl:
@dangunter recently suggested https://pypi.org/project/pyleft/, a tool that checks that type annotations are present without actually enforcing them (as opposed to e.g. MyPy, pyright, ...). If we can manage to make it work on a limited subset of files, this could be a good intermediate step going forward.

PR #1184, opened 2024-01-04
Adding type annotations to SDoE

sotorrio1:
## Fixes/Addresses:

Addresses issue #1111 

## Summary/Motivation:
Adding type annotations to all SDoE files

- adding SDoE type annotations
- adding some function descriptions 

### Legal Acknowledgement

### DIFF
--- a/foqus_lib/framework/sdoe/df_utils.py
+++ b/foqus_lib/framework/sdoe/df_utils.py
@@ -12,20 +12,30 @@
 # respectively. This file is also available online at the URL
 # "https://github.com/CCSI-Toolset/FOQUS".
 #################################################################################
+from typing import Optional, Union, Tuple
 import pandas as pd
 
 
-def write(fname, df, index=False):
+def write(fname: str, df: pd.DataFrame, index: bool = False):
+    """
+    Writes data frame as csv file
+    args: fname, df, index
+    returns: None, saves df to csv file
+    """
     if index:
         index_label = "__id"
     else:
         index_label = None
-    # write data frame as csv file
+
     df.to_csv(fname, index=index, index_label=index_label)  # do not write row names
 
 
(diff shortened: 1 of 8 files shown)


sotorrio1:
@lbianchi-lbl & @franflame I'm no expert on type annotations but hopefully this is a good starting point.

franflame (dismissed):

lbianchi-lbl (commented):
Thanks @sotorrio1 for doing the work. This looks good as it is. One thing I'd consider changing before merging would be to use `typing.List` and `typing.Dict` (instead of `list` and `dict` respectively[1]) to be able to add information about the type of contained items, e.g. 

```py
from typing import List
...
def split_string(s: str) -> List[str]:
   ...
```

Concretely, this would be useful to distinguish between e.g. `List[int]`, `List[str]`, etc as IIRC both show up in the SDoE function arguments.

---

[1]: This in not needed with Python 3.9+ as the builtin `list`, `dict`, `tuple` etc support the type annotation `[]` directly, but we're currently still maintaining compatibility with Python 3.8.

sotorrio1:
> Thanks @sotorrio1 for doing the work. This looks good as it is. One thing I'd consider changing before merging would be to use `typing.List` and `typing.Dict` (instead of `list` and `dict` respectively) to be able to add information about the type of contained items, e.g.
> 
> ```python
> from typing import List
> ...
> def split_string(s: str) -> List[str]:
>    ...
> ```
> 
> Concretely, this would be useful to distinguish between e.g. `List[int]`, `List[str]`, etc as IIRC both show up in the SDoE function arguments.
@lbianchi-lbl Sounds good! I'll take care of that. Thanks!

lbianchi-lbl (approved):
Let's get this in now, and move any additional changes to future PR(s).

Issue #1196, opened 2024-03-01
Streamline scaling functionality


lbianchi-lbl:
- [ ] Reimplement scaling/unscaling functionality in `foqus_lib.framework.graph.nodeVars` to support numpy/pandas data structures
- [ ] Add testing (potentially/partially property-based, e.g. Hypothesis)
- [ ] Register scaling functions (e.g. as Setuptools entry points) to make them accessible throughout the FOQUS codebase
- [ ] Integrate the scaling functions in the ANN surrogate plugin ensuring that the available options match the registered scaling functions
- [ ] Extend new scaling functionality integration to other parts of the codebase (e.g. ML/AI plugin)

PR #1197, opened 2024-03-01
Add first implementation of scaling functions

franflame:
## Fixes/Addresses:

#1196 

## Summary/Motivation:

Adding initial implementation of scaling functions for ML/AI plugin, with tests included. 

### Legal Acknowledgement


foqus_lib/framework/surrogate/scaling.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    return result
+
+
+def unscale_power2(array_in, lo, hi):
+    result = np.log10(9.0 * array_in / 1.0 + 1) * (hi - lo) + lo
+    return result
+
+
+class BaseScaler:
```
bpaul4, line 141: I may be misunderstanding the usage of the annotations here, but what is the outcome of this class? It seems that arrays that are transformed will raise exceptions for any input.

franflame: I'll add a comment to explain the purpose of the BaseScaler class; transform() and inverse_transform() should be implemented by the derived classes, so it raises an error if called from the base class.

lbianchi-lbl:
This is affected by the Tensorflow version issues that are being worked on in #1223. This should be considered on hold until those issues have been resolved.

PR #1217, opened 2024-04-10
Resolve SMT Method Changes / Support for JENN Models

### LINKED ISSUE #1205: Resolve SMT Method Changes
One of our dependencies, SMT, internally modified it's graph-enhanced neural network (GENN) method calls. Notably, the GENN module no longer exists in version 2.5.0 and has been replaced by the separate package Jenn. We should update our node module and tests to use the new syntax.

bpaul4:
## Fixes/Addresses:

Resolves #1205 by updating SMT GENN syntax to the latest supported version, and by adding usage of JENN (a new dependency of SMT GENN) as well.

### Legal Acknowledgement


examples/other_files/ML_AI_Plugin/mea_column_model_training_jenn.py
```diff
@@ -0,0 +1,151 @@
...
+with open("mea_column_model_jenn.pkl", "wb") as file:
+    pickle.dump(model, file)
+
+# load model as pickle format
+with open("mea_column_model_jenn.pkl", "rb") as file:
+    loaded_model = pickle.load(file)
+
+
+print(y_pred)
+print("SSE = ", SSE)
```
lbianchi-lbl, line 151: Would it be possible to wrap this code in a `main()` function, so that it's not run on import?

```py

def main():
    # move code here


if __name__ == "__main__":
    main()
```

Please disregard if the current structure is required (e.g. the code needs to be at the module scope for the dynamic module import to work correctly).

examples/other_files/ML_AI_Plugin/mea_column_model_training_smtgenn.py
```diff
@@ -30,93 +33,108 @@
...
     pickle.dump(model, file)
 
 # load model as pickle format
-with open("mea_column_model_smt.pkl", "rb") as file:
+with open("mea_column_model_smtgenn.pkl", "rb") as file:
     loaded_model = pickle.load(file)
+
+
+print(y_pred)
+print("SSE = ", SSE)
```
lbianchi-lbl, line 140: See previous comment for using `if __name__ == "__main__":`, if possible.

lbianchi-lbl (approved):
A minor question about the possibility of using `if __name__ == "__main__":` for scripts, otherwise looks good as far as I can tell.

examples/other_files/ML_AI_Plugin/mea_column_model_training_jenn.py
bpaul4: That should be possible for the code that calls the training method, as all of the training examples are only used to generate the model files. The training methods that create the models may need to be on the module level - Keras for sure needs this to register and load the model class object.

Some of the Keras example files are vehicles for the class objects, e.g. `mea_column_model_customnormform.py` must be copied to the working directory during testing to load the registered model class in the plugin code, and those non-training files don't have any "main" code to execute.

I can update that here, or in another PR.

lbianchi-lbl (approved):

Issue #1195, opened 2024-02-27
Intermittent failures in TestOUU.testRunOUU on macOS ARM


lbianchi-lbl:
```log
---------------------------- Captured stdout setup -----------------------------
0
1
2
3
4
5
6
7
8
9
10
11
   variable 1 Y

**********************************************************************
*      Welcome to PSUADE (version 1.10.0)
**********************************************************************
PSUADE - A Problem Solving environment for 
         Uncertainty Analysis and Design Exploration (1.10.0)
(for help, enter <help>)
======================================================================
psuade> readApplication WARNING: opt driver PSUADE_LOCAL not found.
**********************************************************************
PSUADE : Search for local minima in sample space.
         number of samples = 1
	 optimization starting point 1 : 
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
		   0.00000000e+00
				 Y =   1.00000000e+35
**********************************************************************
**********************************************************************
PSUADE OPTIMIZATION 1 (1) : 
	 starting X(     1) =   0.00000000e+00
	 starting X(     2) =   0.00000000e+00
	 starting X(     3) =   0.00000000e+00
	 starting X(     4) =   0.00000000e+00
	 starting X(     5) =   0.00000000e+00
	 starting X(     6) =   0.00000000e+00
	 starting X(     7) =   0.00000000e+00
	 starting X(     8) =   0.00000000e+00
	 starting X(     9) =   0.00000000e+00
	 starting X(    10) =   0.00000000e+00
	 starting X(    11) =   0.00000000e+00
	 starting X(    12) =   0.00000000e+00
	 starting Y =   1.00000000e+35
**********************************************************************
**********************************************************************
*     1- OR 2-STAGE OPTIMIZATION UNDER UNCERTAINTY (OUU)
======================================================================
This optimization capability solves the following problem:

   minimize_Z1 { Phi_{Z3,Z4} [ G(Z1,Z2,Z3,Z4) ] } 

   subject to either:
    (a) bound constraints on Z1, Z2, and Z4 (use bobyqa); or 
    (b) no constraint on Z1, Z2, and Z4 (use newuoa); or
    (c) inequality constraints on Z1,Z2,Z3,Z4 (use cobyla)
[...]

PR #1221, opened 2024-05-09
Removing xfail for TestOUU.testRunOUU

ksbeattie:
## Fixes/Addresses:

Addresses #1195 by removing xfail since I was not able to get it to fail.

## Summary/Motivation:

Can't reproduce this failure on my fancy new MacOS laptop with an M3 Max CPU.

- Remove the `pytest.mark.xfail` for `TestOUU.testRunOUU`

### Legal Acknowledgement


lbianchi-lbl:
I think this is might be the original error showing up again:

image

ksbeattie:
Looks like we should leave xfail in place for now, give that this is looking like a python 3.11/3.12 issue on Apple Silocone

PR #1242, opened 2024-09-24
Foqus cloud

boverhof:
## Fixes/Addresses:

Moves specification of file path locations to foqus.core, including changing default location for SimSinter and TurbineLite.   This is incomplete, but the intent is to provide one place where these paths can be managed, specified and requested.

 improvements to the foqus.service module, which is used by the foqus cloud.  

-
-

### Legal Acknowledgement


foqus_lib/core.py
```diff
@@ -0,0 +1,211 @@
+###############################################################################
+# FOQUS Copyright (c) 2012 - 2021, by the software owners: Oak Ridge Institute
```
ksbeattie, line 2: Make sure this is the new copyright header, with at least this year in the range.

PR #1253, opened 2024-12-11
Remove support for Python 3.8

ksbeattie:
## Fixes/Addresses:

[PyQt5-sip was just updated to 12.16.0](https://pypi.org/project/PyQt5-sip/#history) which removed support for Python 3.8.  This is now breaking the CI for PR #1252 in an entirely unrelated way.

## Summary/Motivation:

Since [Python 3.8 is now EOL](https://devguide.python.org/versions/), we should remove support for it as well.

- Remove python 3.8 from our CI and documentation
-

### Legal Acknowledgement


ksbeattie:
Things I noticed while doing this:
- We should look into updating the versions of the following that we use (in `requirements-dev.txt`):
  - black: from 24.3.0 to 24.10.0
  - pylint: from 3.1.0 to 3.3.2
  - astroid: from 3.1.0 to 3.3.6
- The [Matlab docs](https://foqus.readthedocs.io/en/stable/chapt_matlab/reference/matlab_foqus_intro.html) show support for now very old versions of Python

ksbeattie:
After e11d490d18ce09950c670b46161325c314e187f0 which just fixed the version of PyQt5-sip to 12.15.0 for python 3.8, all tests passed.

.github/workflows/nightlies.yml
```diff
@@ -23,8 +23,8 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - '3.8'
           - '3.9'
+          - '3.10'
```
lbianchi-lbl, line 27: It might make sense to set these to `3.12`, so that the oldest and newest Python version are both tested

requirements-dev.txt
```diff
@@ -6,10 +6,6 @@
 # - also pinning the version for Astroid (used by pylint to analyze the AST)
 #   since there can be significant differences between (non-major) versions,
 #   both in terms of behavior and performance
-# - we need to use Pylint 2 with Python 3.8 b/c of a critical error (stack overflow)
-#   occurring with Pylint 3
-pylint==2.17.7;python_version=="3.8"
-astroid==2.15.8;python_version=="3.8"
 pylint==3.1.0;python_version>"3.8"
 astroid==3.1.0;python_version>"3.8"
```
lbianchi-lbl, line 10: It might make sense to remove the `python_version>3.8`, since that's now always going to be true.
```suggestion
pylint==3.1.0
astroid==3.1.0
```

ksbeattie:
@boverhof & @bpaul4 can I get a quick review of this?  I can then get #1252 working again.

.gitignore
```diff
@@ -53,4 +53,6 @@ gams
 logging.conf
 logs
 turbine.cfg
-user_plugins
\ No newline at end of file
+user_plugins
+turbine_aws.cfg
+user_ml_ai_models/__init__.py
```
bpaul4, line 58: Why was this added here?

ksbeattie: Because I accidentally added them in fa2d5b0e979ef915151bf6078ce25636f2a77272 !  I figured I (or others) might make the same mistake again so I added them to be ignored.

PR #1275, opened 2025-12-11
Fixing typos

ksbeattie:
## Fixes/Addresses:

New batch of typos.

## Summary/Motivation:
Fixed typos and added `indx` as an allowed exception

- Better spelling
-

### Legal Acknowledgement


.typos.toml
```diff
@@ -44,3 +44,4 @@ PyCodeInterupt = "PyCodeInterupt"
 simin_nam = "simin_nam"
 simout_nam = "simout_nam"
 displayAvailablityWarning = "displayAvailablityWarning"
+indx = "indx"
```
bpaul4: Where is this being used?

ksbeattie: There's a bunch of places where a var `indx` is used and typos now wants to change that to `index`.

Take a look at [JB's PR test failures](https://github.com/CCSI-Toolset/FOQUS/actions/runs/20082742668/job/57613458398?pr=1274) for what motivated this PR.

The alternative is to just change all those var to be "index" and make the above exception, I could go either way on that, I just chose the this out of laziness.

ksbeattie:
> Regarding the `jenn` error, it appears that that package recently changed import locations such as the model class `jenn.model.NeuralNet` is now just `jenn.NeuralNet`. Might take a bit of digging to see what the correct way to call the model object is now, but shouldn't be more than a single-line update in one or more places.
> 
> See the recent commits in the last few weeks: https://github.com/shb84/JENN/pulls?q=is%3Apr+is%3Aclosed

heh, a quick fix nested inside of another quick fix!  If you think you can get that figured out, that would be good.  Unless we combine those `jenn` changes into this PR, either this one or yours would need to be force-merged.  Not a problem, just something to be aware of.

